PR #8, opened 2015-12-29
chore(readme): add the link to arch linux aur

actionless:


andlabs:
I will consider merging this once the library gets more stable and gets actual documentation. It now has install rules, so it should be safe to use now, but eh.

actionless:
now?

andlabs:
On the next packaged release, so I can consider the library stable to use (and hopefully other distributions will follow).

actionless:
i was asking because i found few mentions of your lib in few different internet media almost at the same time (about few weeks ago) so i thought it was a movement from your side to announce it

andlabs:
Merging now. How will the AUR choose between stable and unstable releases?

actionless:
usually it's naming convention for AUR package, ie `libui` for stable, `libui-git` for latest git version

andlabs:
All right, thanks! Enjoy using the AUR package.

PR #10, opened 2016-01-01
Makefile changes + Travis file

fclairamb:
Very impressive work you did !

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 
+# This makes us use Ubuntu 14 instead of 12
 dist: trusty
 
 language: cpp
 script:
-  - cat /etc/lsb-release
   - sudo apt-get update
   - sudo apt-get install libgtk-3-dev -y || sudo apt-cache search libgtk3
   - make -f GNUmakefile
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+
+dist: trusty
+
+language: cpp
+script:
+  - cat /etc/lsb-release
+  - sudo apt-get update
+  - sudo apt-get install libgtk-3-dev -y || sudo apt-cache search libgtk3
+  - make -f GNUmakefile
+  - make -f GNUmakefile example EXAMPLE=histogram
+
+# Only working things are added here... Some interesting commands would be:
+#  - make -f GNUmakefile examples
+#  - make -f GNUmakefile test
+#
+# But they both won't work
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,5 @@ script:
   - sudo apt-get update
   - sudo apt-get install libgtk-3-dev -y || sudo apt-cache search libgtk3
   - make -f GNUmakefile
-  - make -f GNUmakefile example EXAMPLE=histogram
-
-# Only working things are added here... Some interesting commands would be:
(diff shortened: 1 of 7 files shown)


andlabs:
When I say `TODO examples rule` I meant a rule like

```
examples: libui
    $(MAKE) example EXAMPLE=controlgallery
    $(MAKE) example EXAMPLE=histogram
    $(MAKE) example EXAMPLE=future-example
```

that builds all the examples for you. I'm not sure if I want to do this.

I wouldn't put the install rule as you have it in `GNUmakefile`; as it stands that will only be correct for Unix systems that are not OS X. The best place would probably be within build/GNUbasegcc.mk, but with some more support in each of the $(OS)/GNUosspecific.mk to clean up the gunk related to sonames/install_names and the location of the version suffix in the filename. That bit is on my mental TODO list. (build/GNUbasemsvc.mk install would just echo that make install is meaningless on Windows, unless there's good reason to imitate the behavior.)

Is there precedent to having `$(prefix)` be all lowercase? If there's existing make-based build systems that use an all-uppercase `$(PREFIX)` I'd go for that instead.

How does the test build not work in the Travis environment? What errors do you get? controlgallery is not yet working right now; I changed the API but not the example.

That particular warning fix is stale code; GTK+ automatically handles resizing controls when necessary for me. No need to change anything here; I'll get rid of that myself later.

fclairamb:
Yes you're right about all of this. I made some changes

### test rule
You have the output here: https://travis-ci.org/fclairamb/libui/builds/99757300

### install rule
I agree except I started simple with some OS specific files. Having variables all around (for names) makes it complex to start and we could actually need OS-specific install rules.

## the warning
It's simply to make sure there's no warning and be able to switch to "-Werror" as soon as possible.

This is just a proposal, you can take whatever you want from this pull request.

andlabs:
Oh right, I forgot to commit page8.c until long after the previous push. Try now.

fclairamb:
> Oh right, I forgot to commit page8.c until long after the previous push. Try now.

As you can see this is fixed : https://travis-ci.org/fclairamb/libui but there are still some warnings.

> [...] the equivalent line on the Unix one would make a symbolic link to the source directory, not to the file that was just copied.

Symbolic links can have destination relative to their path. This is what we have here. On Linux it just takes it as is (without trying to create an absolute path from the local file) and I guess it does the same on Mac Os X.

> Also where does $(DESTDIR) come from?

$DESTDIR is a Makefile variable used to define where to put the binaries when building packages. I know it's mandatory for debian packaging and I guess it is for all other packaging systems too.

If it's not provided, then it's empty and the root is used, which is what you would expect with a manual ```make install```.

andlabs:
Ah, good to know (for both points); thanks. I'll go through this one last time later and merge it (but will probably change a few other things). Thanks again!

fclairamb:
Yes you can change whatever you like. I'm just trying to make it easier for you to validate what you push (through travis) and easier for developers to integrate it (through the install rule).
Anyway the [overall changes](https://github.com/andlabs/libui/pull/10/files) are quite small.

andlabs:
Many years later update: I'm moving the build system to meson, which will handle DESTDIR for us: https://mesonbuild.com/Installing.html#destdir-support

PR #11, opened 2016-01-04
Adding travis icon

fclairamb:
Thank you for accepting my previous pull request. Tiny images like this are always pleasant to see. It shows the developer took the time to make sure it's actually building.


andlabs:
Right; I had set this up on my end but forgot how to add the badge. Thanks.

PR #12, opened 2016-01-13
darwin: Remove warning at initializing NSTabViewItem

hajimehoshi:
This change removes warning at initializing NSTabViewItem.

### DIFF
--- a/darwin/tab.m
+++ b/darwin/tab.m
@@ -93,7 +93,7 @@ void uiTabInsertAt(uiTab *t, const char *name, uintmax_t n, uiControl *child)
 	[t->views insertObject:view atIndex:n];
 	[t->margined insertObject:[NSNumber numberWithInt:0] atIndex:n];
 
-	i = [[NSTabViewItem alloc] initWithIdentifier:nil];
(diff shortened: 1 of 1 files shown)


andlabs:
I'm not sure why clang is complaining about this; the documentation explicitly says that the identifier is optional, but nothing quite says what the consequence of using a `nil` identifier is (if there is any) or of having duplicate identifiers in the same NSTabView (there seems to be none). I think it's a misdeclaration of the method, as the Swift header explicitly does not allow the identifier to be `nil`. Maybe a bug report is in order...

Personally I'd rather continue using designated initializers, which `init` doesn't seem to be (nothing says). Faking a unique identifier (perhaps with a dummy NSObject) might be better.

For what it's worth, NSTableColumn also has this issue and I'm going to have to address that too. At least with that the documentation is more clear on how the identifier field is intended to be used. (And in that case, `initWithIdentifier:` *is* listed as the designated initializer.)

hajimehoshi:
Sure. I prefer passing a dummy object if anything, but I don't have a strong opinion in that regard. Let me close this PR.

PR #13, opened 2016-01-13
Add .gitignore

hajimehoshi:
This change adds .gitignore to ignore 'out' and '.obj' directories.


andlabs:
This PR is no longer relevant because cmake doesn't use `obj` and `.out`, and I doubt the build system I move to, whichever that is, will do the same.

PR #15, opened 2016-01-15
Add CMake build systems

yageek:
As I was experiencing some issues on building it on Windows using the default Makefiles, 
I added files to build the library and test program using CMake.

Tested on OSX El Capitan and on Microsoft Windows 7.
Should be confirmed for Haiku (may need some additions) and Unix (should work).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,20 @@
-os:
-  - linux
-
-# This makes us use Ubuntu 14 instead of 12
-dist: trusty
-
 language: c
+
+matrix:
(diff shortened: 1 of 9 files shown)


andlabs:
Unfortunately, I don't have any intention on changing the build system right now. Sorry.

I would, however, like to know what issues you had with using the Makefiles on Windows. What errors did you get?

yageek:
I close the pull request and create an issue.

andlabs:

PR #18, opened 2016-02-12
CMake support

vivkin:
Build and install library, examples and test with CMake


vivkin:
I'm tested on OS X 10.10 with Clang, Ubuntu 15.10 with GCC 5.2 and Windows 10 with Visual Studio 2015 Community Edition.

andlabs:
As I mentioned in #15, I have no intention of changing the build system right now, especially to one that doesn't come with the OS and default development tools. Sorry. I already am bending this rule with requiring GNU make on Windows.

Perhaps this may change in the future.

Thank you for the contribution in the meantime.

kainjow:
Make on Windows seems far more foreign than CMake, or any other format that generates a Visual Studio project, which I imagine a lot more devs are comfortable with. At least by using CMake you can still using makefiles or an IDE, the choice is up to each. I would strongly suggest you reconsider this.

andlabs:

PR #19, opened 2016-04-04
General Improvements

tcyrus:
* Change Badge from PNG to SVG
* Shorten Stackoverflow URLs
* Improve formatting of `.travis.yml`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,15 +2,21 @@ os:
   - linux
 
 # This makes us use Ubuntu 14 instead of 12
+sudo: required
 dist: trusty
 
 language: c
+
+before_install:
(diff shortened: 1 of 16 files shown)


andlabs:
Why require sudo in the travis file?

tcyrus:
Since the sudo user is used to install packages, I thought it would be a good idea to add `sudo: required` to the `.travis.yml` file.

andlabs:
The source has changed considerably since them; I'll apply the README fixes myself, though. Thanks! (I am not sure if I want to shorten the Stack Overflow URLs, but that's definitely something I can consider; just expect more links as time goes on. Don't bother with TODO.md; that's just notes and isn't really intended to look pretty.)

PR #20, opened 2016-04-18
[mac os x] ensures newly created window is drawn in front and active

anilybba:
I noticed during my testing that when I run my code from a terminal window on Mac OS X and the terminal window has the focus, the newly drawn window ends up behind my terminal. This patch forces the newly drawn window to the foreground when it's drawn, regardless of which app has the focus.

### DIFF
--- a/darwin/window.m
+++ b/darwin/window.m
@@ -100,6 +100,7 @@ static void windowCommitShow(uiControl *c)
 	uiWindow *w = (uiWindow *) c;
 
 	[w->window makeKeyAndOrderFront:w->window];
+	[NSApp activateIgnoringOtherApps:YES];
 }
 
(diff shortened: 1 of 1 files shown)


andlabs:
This will break .app bundles; see [this](http://stackoverflow.com/questions/25318524/what-exactly-should-i-pass-to-nsapp-activateignoringotherapps-to-get-my-appl).

anilybba:
Oh wow, thanks for showing me that, I learned something. I'll abandon this pull request for now and just get in the habit of building app bundles for my projects on os x. Sorry I didn't do more thorough research before covering the same ground you covered in the past!

andlabs:
That's all right; don't feel bad :) The contributions are always welcome.

PR #21, opened 2016-05-15
added: uiMainStep for handling events manually in a blocking and non-blocking fashion

ProPuke:
This hasn't had proper testing. Only used it in linux atm, but think it _should_ be okay in windows and osx (which probably means it isn't, and needs testing).

`uiMainStep(int blocking)` - Run a single iteration of the main loop. If `blocking` is non-zero this may block until the next available event, before processing.

For use in your own manual program loops.


andlabs:
Interesting. I need to do something like this in my various stddialogs.c implementations. What would be useful to the programmer using this library though?

The Cocoa and Windows implementations are wrong; the Cocoa version has some typos and misses some subtleties of `[NSApplication run]`, while the Windows version uses undefined symbols.

ProPuke:
Sorry, what I get for doing those blind >.>

It's useful for any apps where you still want control of the main loop - for instance if you have a game or gamedev tool that uses libui, and you need a tight render-loop and other updates alongside your ui. I guess you could delegate the ui handling to a separate thread, but that seems a bit of a convoluted solution (and needless threads usually cause problems).

In my case I'm using sdl and a few other things alongside libui, so I need to maintain my own event loop.

I cleaned up the windows spam (I'd pasted something for reference and forgot all of it). I had presumed that firing off events manually with [NSApplication nextEventMatchingMask] and [NSApplication sendEvent] was a suitable replacement for [NSApplication run]? I could be totally wrong there. I did kinda guess. Windows code _should_ be fine now.

andlabs:
It is a suitable replacement; it is effectively what NSApplication does. However, windows with modal sheets complicate this. I also think you're supposed to have each call in its own NSAutoreleasePool, and GNUstep also updates the main menu after sending certain types of events.

And yeah I guess you have a point; providing the functionality would be useful. I'm curious how you got SDL and libui to work together!

ProPuke:
Okay, I'm probably not going to have the time to test and fix that in OS X atm 

They don't talk to one-another yet; They just co-exist in the same app at the same time. I will be looking to see if I can embed sdl renderers within libui with `SDL_CreateWindowFrom` next, though.

andlabs:
The whole modal model is a mess, and one I genuinely have no clue how to fix...

pcwalton:
This is also useful for my use case (Servo).

andlabs:
I think I will implement this myself, as there are some code structure changes compared to this I want to make. I am accepting the request, though!

andlabs:
API:

    int uiMainStep(int wait);

If wait is nonzero, wait for a message before processing. If wait is 0, and a message is available, process it immediately (returning either 1 or 0 depending on the message), otherwise return 1.

Returns 1 if a message was processed and 0 if the (current nesting level of the) main loop has quit (for instance, if a `WM_QUIT` is posted on Windows).

andlabs:
All right, I added uiMainStep(). Thanks ProPuke; your code was used as a guide (and in some cases I did have to do some of what you did). I'll leave this PR open for the question I just posed.

andlabs:
Questions moved to TODOs.

andlabs:
Okay, I need to resurrect this.

>It's useful for any apps where you still want control of the main loop - for instance if you have a game or gamedev tool that uses libui, and you need a tight render-loop and other updates alongside your ui. I guess you could delegate the ui handling to a separate thread, but that seems a bit of a convoluted solution (and needless threads usually cause problems).

How do people do this right now on macOS? Because there is no real supported way to do this directly in the same way libui does it now, and even if I did just recreate what `-[NSApplication run]` does internally, that wouldn't be sufficient to replace *all* of `-[NSApplication run]`.

PR #27, opened 2016-05-20
Create barebones example

joaoventura:
A barebones example with a window, vertical box and a label.


kainjow:
IMO, this should be added to the Makefile so it builds with `make examples`.

andlabs:
[I already have a plan for a "first example".](https://github.com/andlabs/ui/wiki/Getting-Started)

PR #45, opened 2016-05-21
fix typo

Extrawurst:
This is supposed to be a C lib right ? ;) even though C++ and C are equally dumb when is comes to enums (enum class excluded)

### DIFF
--- a/ui.h
+++ b/ui.h
@@ -17,7 +17,7 @@ extern "C" {
 #define _UI_EXTERN extern
 #endif
 
-// C++ is really really really really really really dumb about enums, so screw that and just make them anonymous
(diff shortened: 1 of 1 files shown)


andlabs:
No, that comment is accurate. C++ enums are discrete types that cannot be used as bitwise flags (even with an enum class!).

Extrawurst:
What i meant was that ur having a C lib here, so the comment should talk
about C and not C++ right?

andlabs:
While the library is a C library, the C header can be used by C++ code, and someone who writes a C++ program with this library will run into the problems described in the comment (that you can't `&` or `|` enum values without lots of operator overloading). I hope WG21 fixes this stupidity soon.

PR #47, opened 2016-05-22
add bindings

Srdgh:
C#/.Net, D, Haskell, Lua, Rust


mellinoe:
Perhaps just link to the wiki? I see you already listed them there, and this section might get really bloated and out-of-sync if they're listed in both places.

Srdgh:
Makes sense -- done as suggested.

joaoventura:
I included my python bindings in the wiki, pretty barebones by now but since I created a custom-made generator to get the function signatures from the ui.h header file, it is just a matter of filling the gaps for each function (I'm using ctypes for interfacing with C)..

andlabs:
I'll probably wind up using the wiki for documentation somehow instead of the bindings list, but I'll merge it manually.

PR #57, opened 2016-05-24
Enable OS X Travis tests

kainjow:


andlabs:
If this works then thanks for figuring out what I couldn't =P

andlabs:
Cool, it works :D I wonder what I was doing wrong...

PR #59, opened 2016-05-24
Add .gitignore for C projects

parro-it:
I created the .gitignore using [joe](https://github.com/karan/joe).
I find it annoying that every time I build the project the repo git status get dirty...


kainjow:
There is also #13 which I think is all that's currently needed (simpler).

PR #58, opened 2016-05-24
Fix crash when closing program on OS X

kainjow:
This fixes #14. Autorelease pools need to be used to make sure objects get properly released. Unfortunately this produces a new error when menuManager gets deallocated, which I am looking at fixing:

> map.m:25:mapDestroy() POSSIBLE IMPLEMENTATION BUG; CONTACT ANDLABS:
attempt to destroy map with items inside

PR #60, opened 2016-05-24
Fix test crash on exit on OS X

kainjow:
This should fix the crash reported in #58. I believe the issue is related to the menus being released, then the menu items being released afterwards, but still referencing the deallocated menu internally, thus the crash.

Oddly enough, I'm using a CMake project to debug with Xcode and when I set it to a bundle target (so it generates a proper .app), it's harder to reproduce, but when I leave it as a console app, it reproduces every time (same as `make test`).

### DIFF
--- a/darwin/menu.m
+++ b/darwin/menu.m
@@ -32,8 +32,7 @@ static void mapItemReleaser(void *key, void *value)
 	uiMenuItem *item;
  
 	item = (uiMenuItem *)value;
-	// TODO this crashes for me on OS X El Capitan
-//	[item->item release];
+	[item->item release];
 }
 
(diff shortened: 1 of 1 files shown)


andlabs:
Do you see "after uiUninit()" on the debug log in Xcode when you build as an app bundle? Out of curiosity.

kainjow:
Yes I see that log.

SevenBits:
I just built the library on El Capitan and did not experience a crash.

kainjow:
Did you run `make test` and then `./out/test`, then quit the program? That is how I was able to reproduce the crash.

PR #61, opened 2016-05-25
Fix additional OS X memory management issues

kainjow:


darwin/button.m
```diff
@@ -103,7 +103,7 @@ static void defaultOnClicked(uiButton *b, void *data)
 	uiDarwinSetControlFont(b->button, NSRegularControlSize);
 
 	if (buttonDelegate == nil) {
-		buttonDelegate = [buttonDelegateClass new];
+		buttonDelegate = [[buttonDelegateClass new] autorelease];
```
andlabs: What does autoreleasing these change?

darwin/main.m
```diff
@@ -3,6 +3,8 @@
 
 static BOOL canQuit = NO;
 static NSAutoreleasePool *globalPool;
+static applicationClass* app;
```
andlabs: We could get rid of `realNSApp()` with this. I only provide it because `NSApp` by default is `id`, which can confuse clang and generate garbage warnings like `setDelegate:` having the wrong protocol type.

darwin/main.m
```diff
@@ -118,20 +119,26 @@ - (BOOL)applicationShouldTerminateAfterLastWindowClosed:(NSApplication *)app
 		[realNSApp() setMainMenu:[appDelegate().menuManager makeMenubar]];
 
 		setupFontPanel();
-
-		return NULL;
 	}
+
+	globalPool = [[NSAutoreleasePool alloc] init];
```
andlabs: What does moving this down here do?

andlabs:
Nice work in the meantime! It might take a while for me to break the manual refcounting habits I've worked up until now, but maybe these changes will be the beginning of a possible future use of ARC?

darwin/button.m
kainjow: In the above delegates code, clang's analyzer (Xcode > Product > Analyze) will warn:

> libui/darwin/alloc.m:31:3: Incorrect decrement of the reference count of an object that is not owned at this point by the caller

The array retains the objects, and thus becomes the owner. By releasing objects from the array directly, you are releasing an object that doesn't belong to you. You should really only release objects that you hold references to. Since we don't need the object after it's done being stored in the array, we can autorelease the object so that the array has the only reference. Since it doesn't get released until the app quits this is fine.

Autoreleasing should always be preferred over manual releases. One, it's not really more performant to release objects directly. You should use autorelease pools if you need performance (rare). Two, it's one step closer to Automatic Reference Counting, which ideally every Objective-C project should move to, as you'll get slight performance improvements, and better code.

darwin/main.m
kainjow: It was unnecessary to have the global pool created before the local pool, since the local pool covers all autoreleased objects in that function. The global pool should then be created right before exit of uiInit, and released right at the beginning of the uiUnint to catch all autoreleased objects created after uiInit. NSApplication run will create its own pools for each run loop, so we don't need to worry about pools for functions called after uiMain.

Moving this actually uncovered the over-release I found with pageID and view in tabPage.

You could also get rid of NSAutoreleasePool and enclose all uiXxx functions with @autoreleasepool. Could be doable with some type of macro possibly. NSAutoreleasePool is a no-go for ARC, but @autoreleasepools are fine.

kainjow:
Yes it will help get towards ARC support, but this one rule will require a refactor, if it ever comes to it:

    You cannot use object pointers in C structures.
    Rather than using a struct, you can create an Objective-C class to manage the data instead.

andlabs:
Yes, but that refactor isn't as much of a big deal as the fact uiControlHandle() will need to be adjusted somehow...

kainjow:
If you switched from uintptr_t to void* it should be fine. Could use a typedef.

darwin/button.m
andlabs: I will say the goal was not speed, but rather correctness and clarity.

PR #69, opened 2016-05-26
Add support for static linking.

pcwalton:
Use `make STATIC=1` to build a static library.

Some language ecosystems, like those of Rust, have a convention of
static linking in order to make binaries easier to distribute. In those
environments, this feature helps libui fit in more easily.


andlabs:
Will this have namespace collision with internal functions? That was the main reason why I went the shared library route.

pcwalton:
Ah, I see. I'll update the PR when I get a chance to use the `ld -r` techniques for eliminating static symbols.

andlabs:
Ooh, so there is a way to hide those symbols after all? Neat!

HalosGhost:
@sevenbits, seems like this may be what #72 was hoping for.

SevenBits:
@HalosGhost, yes, I'm not sure if there are any edge cases in this PR but I'm hoping that it will be merged.

andlabs:
I won't know about any edge cases until they're tried :) I already explained why I didn't go the static route originally.

Note that on Linux your users are going to need GTK+ installed by the system; *it* does not fully support static linking. That shouldn't be a problem for libui, and GTK+ should be installed on most systems by default anyway.

SevenBits:
I am aware that GTK does not support static linking, but they are likely to have that installed on their system. libui, not so much.

pcwalton:
@andlabs I've updated the PR with `ld -r` support, tested on Mac and Linux. I've verified via `nm` that private symbols no longer leak into the `.a` library. Let me know your thoughts :)

andlabs:
Neato. So can you explain how all this works? I'm genuinely curious, since I wasn't able to deduce this myself earlier. (Also what's the point of the duplicate `$(STATICLIB)` variable?

pcwalton:
Because a `.a` archive is just a series of `.o` files, there's no concept of a hidden symbol local to a static library, and that's why you can't just get `ar` to make your hidden symbols local. But what you can do is to first link all the individual `.o` files onto an intermediate `.o` file, make the hidden symbols in *that* `.o` file local, then create a `.a` consisting of that single `.o` file. On Linux this is done with `objcopy --localize-hidden`, which is presumably for exactly this purpose. On Mac there is no convenient `--localize-hidden` flag that you can pass to `otool` (the Mac equivalent to `objcopy`), but you can pass an `--exported_symbols_list FILE` option specifying a file that lists the individual symbols to be exported. We can generate this file by piping the output of `nm` through a simple `sed` script.

The purpose of `$(STATICLIB)` is to specify that we're actually in the process of building the library itself. With that flag set, the `ar` magic is invoked. The reason why we can't just use `$(STATIC)` there is that the linking Makefile rule is also used by the examples and so forth, and we don't want to invoke `ar` on those.

andlabs:
Ah, understood. Here goes! Thanks again for all this :D

Issue #30, opened 2016-05-20
Qt backend?


flying-sheep:
on my system, GTK isn’t the native toolkit

Chaz6:
I would like a Qt backend also, thanks!

Abscissa:
Was just going to file this, but looks like it's already been posted. Problem is, this...

"...uses the native GUI technologies of each platform it supports"

...and this...

"Unix: GTK+ 3.4 or newer"

...directly contradict each other. GTK is only native on *some* Linux systems. On many, native is Qt. Many Linux users can't stand GTK software, it's just as bad for them as non-native.

andlabs:
GTK+ and Qt are equally native in that a) most modern Linux programs either use one or the other b) there are ways to make each look nearly identical to each other for many common themes.

Writing a Qt backend seems pointless; libui already aims to do what Qt does, so what's the point of providing a less powerful frontend in C to Qt? Just use that directly instead. (While it's true that GTK+ also does this to some extent, it's not as important of a goal for GTK+ as it is for Qt.) GTK+ is also a lot lighter and is more likely to be present than Qt is...

FraGag:
I'm using Plasma 5 along with the Breeze theme, which is primarily a Qt theme. The theme has also been ported to GTK2 and GTK3, but the GTK3 version (I have breeze-gtk 5.6.4) has been broken since the release of gtk3 3.20.x, so I use the default GTK3 theme instead, which is OK but definitely distinct from Breeze, making GTK3 applications feel nonnative.

By offering a Qt backend, libui-based applications would use the user's Qt theme. While it's true that Qt also uses the native look and feel on platforms that have one, it's also true that Qt *defines* the native look and feel on other platforms (especially Qt-based Linux distributions).

Now, having two available backends on Linux poses an interesting architectural challenge. What backend(s) should the built library use? Should the library select the backend at runtime, or should it be selected at compile time? If the latter, should it be possible to install both `libui-gtk3` and `libui-qt5` system-wide? IMO, I'd go with having distinct `libui-gtk3` and `libui-qt5` libraries, which would be provided as separate packages by distributions with appropriate dependencies, as well as a virtual `libui` package (with two options, `libui-gtk3-default` and `libui-qt5-default` or something like that) that just installs a symlink to the selected implementation. Applications that don't depend on the backend would link to `libui`, whereas applications that depend on a specific backend would link to the specific library.

flying-sheep:
[...]

PR #80, opened 2016-05-29
Qt5 backend for libui.

kristjanbb:
Inspired by #30 discussion.

[Screenshots](http://imgur.com/a/ei9rp)

Known issues:
- Mouse and Key event translations are incomplete
- Margin chaning is not implemented
- uiBox controls can not be re-parented (see comment in qt5/tab.cpp)
- It is bit ugly since margins have not been folded..
  ..or any styling to make it look decent for that matter
- Most signals are not supressed on programmatic changes

TODOs:
- Finish unused drawmatrix function
- Fix the uiBox so every uiControl is backed by a QWidget to fix/simplify
  reparenting/margins
- Create minimal [qss](http://doc.qt.io/qt-5/stylesheet.html) once the margins functionality is sorted
- Sum up some libui API thoughts (should have written it down).

I won't fault you for not wanting to support another backend. If you want to integrate this I can polish it next weekend. However I don't have time to maintain at the rate your are developing, but I can anwser any questions anytime.

### DIFF
--- a/qt5/GNUfiles.mk
+++ b/qt5/GNUfiles.mk
@@ -1,0 +1,74 @@
+# 28 may 2016
+
+CXXFILES += \
+	qt5/alloc.cpp \
+	qt5/area.cpp \
+	qt5/box.cpp \
+	qt5/button.cpp \
+	qt5/checkbox.cpp \
+
+#	qt5/child.cpp \
+
+CXXFILES += \
+	qt5/colorbutton.cpp \
+	qt5/combobox.cpp \
+	qt5/control.cpp \
+	qt5/datetimepicker.cpp \
+	qt5/debug.cpp \
+	qt5/draw.cpp \
+	qt5/drawmatrix.cpp \
+	qt5/drawpath.cpp \
+	qt5/drawtext.cpp \
+	qt5/editablecombo.cpp \
+	qt5/entry.cpp \
+	qt5/fontbutton.cpp \
+	qt5/group.cpp \
+	qt5/label.cpp \
+	qt5/main.cpp \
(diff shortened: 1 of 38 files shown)


kristjanbb:
Build with `make OS=qt5` or `make OS=qt5 -j8 examples test` if you want to build everything (examples and test) in one go.

You'll probably need qtbase5-dev (debian/ubuntu). It might fail with old versions than 5.4, but nothing that shouldn't be hard too fix.

waddlesplash:
Very nice. Looks much nicer than GTK backend does on KDE :)

DemiMarie:
I hope this is merged.  libui claims to use the native toolkit for each platform.  QT _is_ the native toolkit on KDE and QT for embedded Linux, just like GTK is the native toolkit on GNOME.

I don't think that QT does everything libui does.  For one, libui is in C and is trivial to bind to, whereas QT is in C++ and very hard to bind to.  Furthermore, QT does not necessarily act native on Windows or OS X — it tries, but does not always succeed.  Finally, QT is a heavy dependency, whereas libui is a much smaller dependency.

Chaz6:
Wow thank you @kristjanbb for the effort you have put in 👍

mischnic:
I have updated this: https://github.com/mischnic/libui/tree/qt5.
Large parts that are still missing: 
- proper grid support / layout fixes
- the font/attributed string api
- fine tuning (or rather coarse) of margins/spacing.

I don't think Qt will ever get merged because 3 GUI toolkits are already enough work (maybe as an *out-of-tree backends*):
#327:
> **Qt**
> Status: unplanned

PR #85, opened 2016-05-30
ondestroy handler, more type conversion macros

gnarz:
ondestroy handler for controls, for language integration use.

### DIFF
--- a/common/GNUfiles.mk
+++ b/common/GNUfiles.mk
@@ -6,6 +6,7 @@ CFILES += \
 	common/debug.c \
 	common/matrix.c \
 	common/shouldquit.c \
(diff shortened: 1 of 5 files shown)


gnarz:
failed because of a problem with the MacOS X build, for which there is also a pull request (#86).

andlabs:
Unfortunately this isn't really going to work. An `OnDestroy()` handler should be done at the control level, not globally.

The conversion macros you provided, except for the uiAreaHandler one (which I might ad myself), are unnecessary because none of those types can pose for anything other than themselves (they are fully opaque).

gnarz:
it works quite well. I'll continue to use it until you provide something at control level.

PR #87, opened 2016-05-31
Add git config.

billyquith:
.gitattributes avoids cross-platform text formatting issues, e.g. line endings.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+* text=auto
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,118 @@
+
+### CMake ###
+
+CMakeCache.txt
+CMakeFiles
+CMakeScripts
+Makefile
+cmake_install.cmake
+install_manifest.txt
+CTestTestfile.cmake
+
+### Python ###
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*.class
+
+# virtualenv
+venv/
+ENV/
+
+### C/C++ ###
+
+# Compiled Object files
+*.slo
+*.lo
+*.o
+*.obj
+*.ko
+*.elf
+
+# Precompiled Headers
+*.gch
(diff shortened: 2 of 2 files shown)


andlabs:
The gitignore will not be necessary, especially not one with all that in it.

What does `auto` do in the gitattributes?

billyquith:
It is pretty much my default one and then I strip it down. If you are using a cross platform build system it will need a fair few things in.

https://www.git-scm.com/docs/gitattributes

> **Set to string value "auto"**
When text is set to "auto", the path is marked for automatic end-of-line normalization. If Git decides that the content is text, its line endings are normalized to LF on checkin.

To be safe you can do: https://github.com/billyquith/ponder/blob/master/.gitattributes

andlabs:
Oh good, that is the line feed behavior I want then :) git should consider using less confusing names (force-LF, force-CRLF, use-native)

billyquith:
It basically modifies files when you check them out and then normalises them when you put them back in the repo. Without this it adds them as-is, regardless of platform, so Windows and UNIX users are clashing.

You might also add .editorconfig for tabs etc but not everyone uses this. 

Personally I prefer spaces as tabs screw up Github code view and https://www.jwz.org/doc/tabs-vs-spaces.html

aoloe:
not sure it's useful to add the files generated by cmake as .gitignore.

afaik, one is supposed to `mkdir build/ && cd build/ && cmake ..`... and, then just ignore `build/` herself...

billyquith:
What harm does it do having them there?

billyquith:
Besides the fact that `build/` is already used, what if you have multiple build configs?

aoloe:
if you have multiple build configs, than you would have to visually ignore multiple build directories :-)

there is no harm directly to longer gitignores.
but i prefer if the projects maintainers organize the generated files in a way that they are in one (or very few) directories that are easy to ignore.

PR #88, opened 2016-05-31
const correctness

billyquith:
I'm working on a C++ wrapper and running into some const issues. It is not a bad idea to make the C code const correct. E.g. visible? and enabled? can be const.

### DIFF
--- a/common/control.c
+++ b/common/control.c
@@ -27,7 +27,7 @@ int uiControlToplevel(uiControl *c)
 	return (*(c->Toplevel))(c);
 }
 
-int uiControlVisible(uiControl *c)
+int uiControlVisible(const uiControl *c)
 {
 	return (*(c->Visible))(c);
 }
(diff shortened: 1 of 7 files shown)


pcwalton:
Improving the const correctness of `libui` also makes the Rust bindings cleaner, since Rust is relatively strict about const correctness. (So far I've been working around it by casting away the const, but it'd be nice to not have to do that.)

andlabs:
Can you two show me how not having the const there breaks things? It's been a while since I've done that kind of thing with C++ and I haven't used Rust, so I'm not sure exactly what happens. I'll merge this after the build system switch.

pcwalton:
This particular PR doesn't affect Rust code. Some other const correctness issues do, though; for example, `uiDrawStroke()` takes a mutable Brush when it should take a const Brush, causing me to have to cast away the const in Rust: https://github.com/pcwalton/libui-rs/blob/master/ui/src/draw.rs#L39

andlabs:
Ah, I see now; thanks. I should probably go through all of ui.h to make it const-correct sometime soon, yes.

billyquith:
These are just the const issues I encountered so far.

PR #97, opened 2016-06-02
CMake improvements & warnings fixes

billyquith:
Fixed a couple of C warnings: switch unused options and unused code.

Couple of Cmake improvements.


billyquith:
Branched from wrong place. Bah.

PR #98, opened 2016-06-02
Cmake improvements

billyquith:
Naming. Logic.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 cmake_minimum_required(VERSION 2.8.12)
 
 # set up our configurations
-set(CMAKE_CONFIGURATION_TYPES Debug Static Release ReleaseStatic)
+set(CMAKE_CONFIGURATION_TYPES Debug DebugStatic Release ReleaseStatic)
 # we load the variables after calling project()
 # default to Debug if no configuration specified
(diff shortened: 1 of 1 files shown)


andlabs:
Declaring `project()` at the top is wrong, as there are several  properties there that must be set before `project()`. Note the comments.

andlabs:
As for the shared library stuff, @kainjow was talking about something built into cmake that's a much better way to handle that than what I am doing now...

billyquith:
> Declaring project() at the top is wrong, as there are several properties there that must be set before project(). 

Where did you read this?

billyquith:
Travis install failed for this, not the changes.

```
W: There is no public key available for the following key IDs:
1397BC53640DB551
W: Failed to fetch https://apt.dockerproject.org/repo/dists/ubuntu-trusty/main/binary-amd64/Packages  Hash Sum mismatch
E: Some index files failed to download. They have been ignored, or old ones used instead.
```

I haven't seen any CMake projects where the main project isn't declared first. The project, according to their docs, just sets a name. It is the scoping that is important, not the order. Whatever exists in the scope at the end of the file is baked into the targets AFAIK.

andlabs:
This PR is no longer relevant due to the CMake files having been rewritten since it was first proposed. Thanks though!

PR #114, opened 2016-06-07
Cmake 2.8.12 doesn't understand LANGUAGES keyword on project command

hugows:
Fortunately we can just use project(libui) since C and CPP are the languages enabled by default.


andlabs:
I wish cmake was more aggressive about using features newer than your minimum version :|

hugows:
Thanks. I actually have zero cmake *and* libui experience - just trying to compile on Ubuntu 14.04 because I'm keeping an eye on this library. And the error was quite confusing...

PR #122, opened 2016-06-11
Generate OS X app bundles

kainjow:
This is the standard format for OS X applications. Plus when debugging with Xcode it causes the application to become the frontmost process properly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,7 @@ endif()
 
 project(libui)
 option(BUILD_SHARED_LIBS "Whether to build libui as a shared library or a static library" ON)
+option(MAKE_MACOS_BUNDLES "Whether to build test and examples as application bundles" OFF)
 
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/out")
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/out")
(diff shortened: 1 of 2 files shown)


andlabs:
This should be an option.

How would bundles work for the library? Would it be a .dylib included in the Contents/MacOS folder?

kainjow:
Why would it be an option? GUIs should be application bundles, that's how OS X does it. Any other way just seems wrong to me.

The library isn't involved with the bundle. If i were distributing a libui Mac app I would make it static, that would be much simpler distribution.

pcwalton:
It needs to be an option for us at least in Servo, because Cargo doesn't know how to generate `.app` bundles yet. I'd be happy to contribute fixes to make the app frontmost and such--this is definitely doable without a bundle.

kainjow:
It's only for the test and examples, I don't think it affects building the library.

andlabs:
Why kill the branch? I still think it should be an option for the examples; I run the test directly from the command line so

kainjow:
You can still run app bundles from the command line, just run the .app/Contents/MacOS/xxx file.

kainjow:
I'll add an option and default it to OFF, then everyone's happy :)

andlabs:
Revisiting this.

Turns out being able to use Dictation relies on an app bundle. :|

Might as well make it unconditional, maybe, but I'm not sure how that will affect developmers using this library...

andlabs:
Thinking about things like this a bit more, I've been leaning toward making a resource generator tool that is part of libui but run separately, because these are things that can't be done in the program's source itself.

Still not sure if I want the examples or tester to be .app bundles; leaning toward "no" for the tester.

PR #123, opened 2016-06-11
Fix void* warning

kainjow:
"Format specifies type 'void *' but the argument has type 'uiWindow *' (aka 'struct uiWindow *')"

### DIFF
--- a/test/main.c
+++ b/test/main.c
@@ -82,7 +82,7 @@ int main(int argc, char *argv[])
 
 	w = newWindow("Main Window", 320, 240, 1);
 	uiWindowOnClosing(w, onClosing, NULL);
-	printf("main window %p\n", w);
+	printf("main window %p\n", (void*)w);
 
 	uiOnShouldQuit(onShouldQuit, w);
 
--- a/test/menus.c
+++ b/test/menus.c
@@ -47,7 +47,7 @@ static void forceOff(uiMenuItem *item, uiWindow *w, void *data)
 
 static void whatWindow(uiMenuItem *item, uiWindow *w, void *data)
 {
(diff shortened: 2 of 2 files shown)


andlabs:
There are more such warnings elsewhere; I'm not even sure why it happens o_O

HalosGhost:
Because ``printf()``'s ``%p`` specifier expects a ``void *``:

From ``printf(3)``:

>        p      The void * pointer argument is printed in hexadecimal (as if by %#x or %#lx).

This is probably mostly harmless, but doing the explicit casts to avoid the warning output should cause no issues; so if the goal is a completely clean build, this should be replicated anywhere that does something similar.

kainjow:
I only fixed the warnings I saw, which was from clang on OS X, so I'm sure there are other similar issues on the other platform code.

andlabs:
Right, I mean why other pointer types aren't treated as `void *` like they would in any other context in C. I can fix this, but I'd like to know why this is a problem in the first place...

andlabs:
Why kill the branch? I was going to apply it

kainjow:
Hmm, I thought GitHub would retain it for the pull request. Are you not able to merge anymore?

andlabs:
@kainjow no, github disables reopening the PR if the branch is deleted.

----

@HalosGhost and I think this might be strict format argument checking at play. I'll ask the compiler guys...

kainjow:
Interesting discussion about it here:

https://www.quora.com/Is-it-undefined-behavior-if-I-dont-use-void*-when-I-use-p-format-specifier-in-C#!n=12

andlabs:
Yep, that explains it; thanks!

PR #131, opened 2016-06-13
set/get size and position; center and fullscreen; unix/win only

henkman:
So here is the stuff we talked about in code form. As I said I can deliver no darwin implementation.
Also this does not handle the issue with the taskbar on GTK

the test I used is here: https://gist.github.com/henkman/2560aef833bad1f21dfeeed3dac440af

### DIFF
--- a/darwin/window.m
+++ b/darwin/window.m
@@ -275,3 +275,38 @@ uiWindow *windowFromNSWindow(NSWindow *w)
 		return NULL;
 	return [windowDelegate lookupWindow:w];
 }
+
+void uiWindowSize(uiWindow *w, int *width, int *height)
+{
+	// TODO: implement
+}
+
+void uiWindowSetSize(uiWindow *w, int width, int height)
+{
+	// TODO: implement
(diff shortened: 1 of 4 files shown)


andlabs:
Thank you for the effort, but there are several problems

1. Window sizes should refer to the size of the content area, which is the part that does not include titlebars or menubars; this is easier to do on Windows and OS X than on GTK+, but getting it on GTK+ should be easy (hint: `gtk_widget_get_allocation()`)

2. Centering should probably use the monitor work area, not the entire size of the screen; I'm not sure if this is correct though (will need to check what OS X does and what GTK+ suggests)

3. You did not read the The Old New Thing blogposts about fullscreen that I linked, so the implementation of that on Windows is wrong

I'll probably wind up writing this myself, since some of the above details are very subtle...

henkman:
Ok, no problem. If it works in the near future I am a happy camper. That means I can concentrate on my actual project I wanted to use this ui lib for

PR #134, opened 2016-06-14
fix darwin build on travis

waghanza:
use a **newer** *version* of `cmake`
as specified https://gist.github.com/winterz/10cc2741d466cbd8ff21


waghanza:
finally specify option in `CMakeLists.txt` worked
but this has been broken **before**

andlabs:
I'd like an option that works with the minimum version I specified so that Ubuntu 14.04 and CentOS users don't need to install cmake manually. Unless there's a way to specify minimum version in an OS-specific way?

waghanza:
I have remade a real PR, sorry for spamming ;-)

PR #139, opened 2016-06-14
fix osx build

waghanza:
use `objective c++` flag on **os x**


andlabs:
The problem with this is that the code is Objective-C, not Objective-C++, and I'd rather not let legal C++ that is illegal C fly through. There has to be a better way...

andlabs:
```
```

Issue #135, opened 2016-06-14
Add uiFormDelete and uiGridDelete


emersion:
These are missing.

andlabs:
uiFormDelete() was added.

uiGridDelete() is now #309.

PR #141, opened 2016-06-14
Adds uiFormDelete()

emersion:
See #135

### DIFF
--- a/darwin/form.m
+++ b/darwin/form.m
@@ -485,6 +485,11 @@ void uiFormAppend(uiForm *f, const char *label, uiControl *c, int stretchy)
 	[f->view append:toNSString(label) c:c stretchy:stretchy];
 }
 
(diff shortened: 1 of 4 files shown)


andlabs:
If you take a closer look, you'll see that the OS X implementation of `-[formView delete:]` doesn't exist

emersion:
Well, that's the only platform on which I didn't tests because I haven't any OS X device :(

emersion:
Okay, just tried to add `-[formView delete:]`. No idea if it works though.

emersion:
Just updated darwin. I pulled changes you made yesterday on master and tested it on a real macOS device.

andlabs:
Cool, thanks. I'll read through this again in a bit and merge it (making fixes as necessary; again, very subtle stuff). Good work!

PR #143, opened 2016-06-15
Fix Travis builds on osx

tbodt:
I just had it update from cmake 3.0 to cmake 3.5. Now it works fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,15 @@ os:
 # This makes us use Ubuntu 14 instead of 12
 dist: trusty
 
-language: c
-script:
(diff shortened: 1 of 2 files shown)


andlabs:
So there's no way to coerce cmake 2.8.11 to treat Objective-C files as C? That's going to be a problem then. Is there a way to specify a minimum cmake version per platform?

kainjow:
I think there is, I just think the cmake file needs more work and tested directly with 2.8.11 to ensure it actually works. I can look into this soonish.

andlabs:
```
```

Issue #148, opened 2016-06-16
target_compile_options requires CMake 2.8.12+


x64architecture:
CMake function target_compile_options requires CMake 2.8.12+ and libui targets 2.8.11 so it should be worked around or the minimum cmake version should be incremented.

andlabs:
```
[09:49:44]  <+ngladitz>	then you probably hit https://cmake.org/gitweb?p=cmake.git;a=commit;h=95cdf132489c79e88a10fdf7a7566fa002c7680b
[09:50:16]  <+ngladitz>	the fix in that commit is in cmake since v3.1.0
```
I'll probably just bump the minimum required cmake version to that.

ghaberek:
I am still using Ubuntu 14.04, and I found this AskUbuntu answer helpful to get from CMake 2.8 to 3.2.

http://askubuntu.com/a/610352

PR #149, opened 2016-06-16
Fix OS X build

x64architecture:
This pull request fixes the build by setting all of the files with the ".m" file extension flags to the CMAKE_C_FLAGS variable. Personally I think there is no other simple way as support for Objective-C was officially added to CMake somewhere in 2014 (can't find which version).

Tested with CMake 2.8.12 (see issue #148 why I used 2.8.12) and 3.5.2.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,6 +130,16 @@ endmacro()
 add_subdirectory("common")
 add_subdirectory("${_OSNAME}")
 add_library(${_LIBUINAME} ${_LIBUI_SOURCES})
+# hack to allow building Objective-C files with old versions of cmake
+foreach(SRC ${_LIBUI_SOURCES})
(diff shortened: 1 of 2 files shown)


andlabs:
So wait, it was Objective-C support outright that was missing, not just some bug? I feel like I should have known that before but now I'm definitely aware of it, and knowing that I feel bad for rejecting the earlier fixes :|

In fact, hell, maybe I should bump the minimum version. For Ubuntu I can tell people who still use Trusty to use a PPA. Will need to see about CentOS, which is why the requirement is 2.8.11 in the first place.

x64architecture:
[Heres a link](http://thread.gmane.org/gmane.comp.programming.tools.cmake.devel/9371) to the discussion on the cmake mailing list on implementing Objective-C support.

andlabs:
```
```

PR #153, opened 2016-06-17
Adds uiGridDelete()

emersion:
- [x] unix
- [ ] darwin
- [x] windows


andlabs:
The indices that you pas into uiGridDelete() are essentially arbitrary. I'm not sure if I want to expose a Delete function that takes these indices, that takes a uiControl, or that takes a row/column pair... but if I do choose one that takes indices, I'd need a way to get indices out as well. Keep this PR around anyway; the internal implementations are going to need it as they would all have to convert to array indices in the end.

emersion:
I personally would prefer a delete function that takes a uiControl as parameter. We could modify ui*Delete functions to have the same behaviour. A uiGridControlAt(x, y) could be useful. What do you think?

PR #158, opened 2016-06-19
Add test for Xcode

kainjow:


.travis.yml
```diff
@@ -22,3 +22,8 @@ script:
   - rm -rf *
   - cmake .. -G "Unix Makefiles" -DBUILD_SHARED_LIBS=OFF
   - make tester examples
+  - if [ "$TRAVIS_OS_NAME" == "osx" ]; then
+      rm -rf *;
```
andlabs: Will this delete any dotfiles added to the `build/` directory?

andlabs:
This will interfere with binary releases; I'm not sure what to do about that.

mischnic:
I don't see how using the Xcode backend for CMake makes any difference. There is no need to "test" that Xcode can build libui.

andlabs:
Actually that test would be nice, because there are a good number of people who want to build libui using IDE projects instead of makefiles or Ninja files, and the Meson setup will make this much easier too since I don't have any weird makefile-specific hacks like I did with cmake. However, given recent events I am strongly considering replacing Travis with Azure Pipelines, and I want to at least try it first.

andlabs:
The maintainer of meson said the Xcode project generator is still experimental, so meh. Ultimately I decided the extra CI effort (== wait when releasing a version) isn't really worth it; sorry.

Issue #52, opened 2016-05-23
Helper function to open URL?


kainjow:
Looks like this can be done with all supported platforms:

- [gtk_show_uri](https://developer.gnome.org/gtk3/stable/gtk3-Filesystem-utilities.html#gtk-show-uri)
- [-\[NSWorkspace openURL:\]](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSWorkspace_Class/#//apple_ref/occ/instm/NSWorkspace/openURL:)
- [ShellExecute](https://support.microsoft.com/en-us/kb/224816)

andlabs:
I've been thinking about that, yes. (Also ShellExecuteEx is preferable to ShellExecute; the latter has poor error reporting.)

andlabs:
I should probably also mention I am thinking about this in relation to a hyperlink control. Not sure if one is essential enough for the main library.

SevenBits:
@andlabs, it would probably be a good idea to stick to controls that exist natively on all targeted platforms; this way you don't have to get into the custom controls business. And I know in particular that Cocoa (OS X) does not have a native link control.

andlabs:
There is already one custom control (uiDateTimePicker on GTK+, though it's built out of pre-existing controls). A hyperlink control in Cocoa should be trivial to implement (maybe it's as easy as NSAttributedString?).

PR #160, opened 2016-06-19
Add uiOpenURL()

kainjow:
Tested with OS X 10.11, Ubuntu 16.04, and Windows 10.

Implements #52

### DIFF
--- a/darwin/util.m
+++ b/darwin/util.m
@@ -13,3 +13,8 @@ void disableAutocorrect(NSTextView *tv)
 	[tv setAutomaticLinkDetectionEnabled:NO];
 	[tv setSmartInsertDeleteEnabled:NO];
 }
+
+int uiOpenURL(const char *url)
+{
+	return [[NSWorkspace sharedWorkspace] openURL:[NSURL URLWithString:toNSString(url)]] == YES;
+}
--- a/test/menus.c
+++ b/test/menus.c
@@ -4,6 +4,7 @@
 uiMenu *fileMenu;
 uiMenuItem *newItem;
 uiMenuItem *openItem;
+uiMenuItem *openURLItem;
 uiMenuItem *shouldQuitItem;
 uiMenuItem *quitItem;
 uiMenu *editMenu;
@@ -50,11 +51,21 @@ static void whatWindow(uiMenuItem *item, uiWindow *w, void *data)
 	printf("menu item clicked on window %p\n", w);
 }
 
+static void openURLTest(uiMenuItem *item, uiWindow *w, void *data)
+{
+	if (!uiOpenURL("https://github.com/andlabs/libui/")) {
+		printf("Failed to open url\n");
+	}
+}
+
 void initMenus(void)
 {
 	fileMenu = uiNewMenu("File");
 	newItem = uiMenuAppendItem(fileMenu, "New");
 	openItem = uiMenuAppendItem(fileMenu, "Open");
+	uiMenuAppendSeparator(fileMenu);
+	openURLItem = uiMenuAppendItem(fileMenu, "Open URL");
+	uiMenuItemOnClicked(openURLItem, openURLTest, NULL);
 	uiMenuAppendSeparator(fileMenu);
 	shouldQuitItem = uiMenuAppendCheckItem(fileMenu, "Should Quit");
 	quitItem = uiMenuAppendQuitItem(fileMenu);
--- a/ui.h
+++ b/ui.h
@@ -665,6 +665,8 @@ _UI_EXTERN int uiGridPadded(uiGrid *g);
 _UI_EXTERN void uiGridSetPadded(uiGrid *g, int padded);
 _UI_EXTERN uiGrid *uiNewGrid(void);
 
+_UI_EXTERN int uiOpenURL(const char *url);
+
 #ifdef __cplusplus
 }
 #endif
--- a/unix/util.c
+++ b/unix/util.c
@@ -8,3 +8,8 @@ void setMargined(GtkContainer *c, int margined)
 	else
 		gtk_container_set_border_width(c, 0);
 }
+
+int uiOpenURL(const char *url)
+{
+	return gtk_show_uri(NULL, url, GDK_CURRENT_TIME, NULL) == TRUE;
+}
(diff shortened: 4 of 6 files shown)


andlabs:
What is the use of the return value?

kainjow:
Boolean, whether the URL was opened or not. For example if you give it a URL that isn't handled by any program, it should fail.

andlabs:
All right, thanks.

For future reference (you don't have to do it here; I'll change it myself) the convention I use is to compare to the false value rather than one of the many true values. I wonder if I should even bother with the explicit checks; they're mainly to highlight that we have different boolean types (and I might add a uiBool just to be safe because I dislike stdbool).

kainjow:
+1 on uiBool

andlabs:
As for the return value, my concern now is whether the OS will display a message on our behalf if the URL failed to open or not. Will it?

kainjow:
On Windows 10 if I make the protocol "httpss" to force it to fail, it actually didn't return false, and it showed this dialog:

windialog

Even if I set `info.fMask` to `SEE_MASK_FLAG_NO_UI` it still shows up. I noticed Qt is using the older `ShellExecute` and when I use that function the same way that same Windows dialog still pops up.

OS X shows something similar, but it will return false:

mac

GTK will return false without anything else showing.

andlabs:
Right; what I didn't want to happen was two error boxes showing up, one from the OS and one from us, both saying slightly different versions of the same thing. How to resolve this I'm not sure...

kainjow:
Well we could just change it to not return anything, so either it happens or it doesn't, and then the OS may or may not show something.

andlabs:
That could be an option, yes. But let's see if it's necessary, I suppose...

I also wonder what other things can be opened with this function, apart from absolute URLs with a schema. Using the `open` verb on ShellExecuteEx() lets you open anything specified on the standard Windows Run dialog, for instance. If the other platforms have the same functionality, then we'll probably need to think of a new name for this function.

andlabs:

darwin/util.m
```diff
@@ -13,3 +13,8 @@ void disableAutocorrect(NSTextView *tv)
 	[tv setAutomaticLinkDetectionEnabled:NO];
 	[tv setSmartInsertDeleteEnabled:NO];
 }
+
+int uiOpenURL(const char *url)
+{
+	return [[NSWorkspace sharedWorkspace] openURL:[NSURL URLWithString:toNSString(url)]] == YES;
```
andlabs: Again, please change to `!= NO`.

I would also add the following TODOs:

```c
// TODO how should we handle errors, if any? do we need to display any? see https://github.com/andlabs/libui/pull/160#issuecomment-227027962 for more information
// LONGTERM FUTURE when resolving the GTK+ LONGTERM FUTURE about sandboxing, we should consider how this call plays with app sandboxes...
```

test/menus.c
```diff
@@ -50,12 +51,22 @@ static void whatWindow(uiMenuItem *item, uiWindow *w, void *data)
 	printf("menu item clicked on window %p\n", w);
 }
 
+static void openURLTest(uiMenuItem *item, uiWindow *w, void *data)
+{
+	if (!uiOpenURL("https://github.com/andlabs/libui/")) {
```
andlabs: Style consistency nit: remove braces from single-line `if` body

unix/util.c
```diff
@@ -8,3 +8,8 @@ void setMargined(GtkContainer *c, int margined)
 	else
 		gtk_container_set_border_width(c, 0);
 }
+
+int uiOpenURL(const char *url)
+{
+	return gtk_show_uri(NULL, url, GDK_CURRENT_TIME, NULL) == TRUE;
```
andlabs: Again, use `!= FALSE`.

I would add the following comments:

```c
// TODO see if we should present the returned error or not; see https://github.com/andlabs/libui/pull/160#issuecomment-227027962 for more information
// LONGTERM FUTURE GTK+ 3.22 deprecates this in favor of a new gtk_show_uri_on_window() that requires a GtkWindow as a parameter for sandbox purposes (which might also handle the above TODO...)
```

windows/utilwin.cpp
```diff
@@ -74,3 +74,20 @@ void uninitUtilWindow(void)
...
 }
+
+int uiOpenURL(const char *url)
+{
+	SHELLEXECUTEINFOW info;
+	WCHAR *wurl;
+	BOOL ret;
+
+	wurl = toUTF16(url);
+	ZeroMemory(&info, sizeof(info));
```
andlabs: Style consistency nit: space after `sizeof`

windows/utilwin.cpp
```diff
@@ -74,3 +74,20 @@ void uninitUtilWindow(void)
...
+	info.cbSize = sizeof(info);
+	info.lpVerb = L"open";
+	info.lpFile = wurl;
+	info.nShow = SW_SHOWNORMAL;
```
andlabs: I would probably use `SW_SHOWDEFAULT`, but I'm not sure...

windows/utilwin.cpp
```diff
@@ -74,3 +74,20 @@ void uninitUtilWindow(void)
...
+	ret = ShellExecuteExW(&info);
```
andlabs: TODOs worth adding; I can do this myself later:

```c
// TODO SEE_MASK_FLAG_NO_UI? (also see the following TODO)
// TODO should we show any errors returned from this? (we should use GetLastError() if so); see https://github.com/andlabs/libui/pull/160#issuecomment-227027962 for more information
// LONGTERM FUTURE when resolving the GTK+ LONGTERM FUTURE about GtkWindow, we should take advantage of the hwnd option here too, even though IIRC that's only for SEE_MASK_FLAG_NO_UI in this case...
```

windows/utilwin.cpp
```diff
@@ -74,3 +74,20 @@ void uninitUtilWindow(void)
...
+	uiFree(wurl);
+	return ret == TRUE;
```
andlabs: Again, use `!= FALSE` instead.

ui.h
```diff
@@ -665,6 +665,8 @@ _UI_EXTERN int uiGridPadded(uiGrid *g);
 _UI_EXTERN void uiGridSetPadded(uiGrid *g, int padded);
 _UI_EXTERN uiGrid *uiNewGrid(void);
 
+_UI_EXTERN int uiOpenURL(const char *url);
```
andlabs: I would add the following comments:

```c
// TODO figure out how every platform behaves if the URL is not a proper URL with method (that is, "www.google.com" instead of "http://www.google.com")
```

PR #175, opened 2016-07-10
Generate documentation using cldoc

cody271:
**Integration with [cldoc](https://jessevdk.github.io/cldoc/) documentation generator**

Once you have installed cldoc & reconfigured cmake: 
```
$ make docs
```

cldoc-libui1

cldoc-libui2

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -217,3 +217,21 @@ macro(_add_exec _name)
 endmacro()
 add_subdirectory("test")
 add_subdirectory("examples")
+
+set(API_HEADERS ui.h)
+find_program(CLDOC cldoc)
+if(NOT ${CLDOC} STREQUAL "CLDOC-NOTFOUND")
+  add_custom_target(docs COMMAND
+    ${CLDOC} generate
+    ${CMAKE_C_FLAGS}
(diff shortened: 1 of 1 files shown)


andlabs:
This seems neat. I'll have to evaluate it in more detail. Can I integrate outside documentation files with this? I'll be providing a bunch of those as well (you can see the beginnings of them in the docs/ folder, specifically the `.md` files).

cody271:
Yes, cldoc supports outside documentation in Markdown format via the ```--merge``` option as described [here](https://jessevdk.github.io/cldoc/generating.html).

parro-it:
@cody271 I used a [custom js script](https://github.com/parro-it/libui-node/blob/master/examples/docs.js) to generate the doc. 

Since all `libui-node` classes are implemented in c++, cldoc seems a good fit for `libui-node` too.

cody271:
@parro-it Nice, I like purpose-built solutions such as your docs.js script.

So before investing more time into this, I'd like to see that cldoc is packaged well enough and sufficiently cross platform for ```libui```. Please try installing cldoc and let me know what you think. I have only installed it on Ubuntu so far. I will try on Mac OS X next. Someone else will need to try installing it on Windows..

PR #188, opened 2016-08-02
unix: fixes centering window when not visible

### LINKED ISSUE #183: Center window by default
Windows are not centered by default. Calling `Show` and `Center` does not work well because the window is first displayed in a corner of the screen and is then centered.

Provide a way to center the window by default.

emersion:
Fixes #183

### DIFF
--- a/unix/window.c
+++ b/unix/window.c
@@ -168,23 +168,30 @@ void uiWindowSetPosition(uiWindow *w, int x, int y)
 
 void uiWindowCenter(uiWindow *w)
 {
-	gint x, y;
-	GtkAllocation winalloc;
-	GdkWindow *gdkwin;
-	GdkScreen *screen;
-	GdkRectangle workarea;
+	gboolean visible;
+	g_object_get(w->widget, "visible", &visible, NULL);
 
-	gtk_widget_get_allocation(w->widget, &winalloc);
-	gdkwin = gtk_widget_get_window(w->widget);
-	screen = gdk_window_get_screen(gdkwin);
-	gdk_screen_get_monitor_workarea(screen,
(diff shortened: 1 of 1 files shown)


andlabs:
My question is: will this work when the window has already been shown once before and then hidden again, or only before the window is shown the first time?

emersion:
Just checked, this only works when the window is shown the first time. Is there a way to detect if the window has never been shown?

Maybe your centering code works when the window has been shown and is now hidden.

andlabs:
I might have to get rid of this code anyway; see #214. There might be another way to do it...

emersion:
If removed, would it be possible to replace it with another function that hints the platform to center the window when opened the first time?

Windows opening at the bottom left corner on macOS are very annoying.

andlabs:
Possibly. Need to look up the GTK+ bits.

I could also have the windows always centered on OS X only (since that's the only one that doesn't have a "just decide a default position for me" option).

emersion:
Okay, just found out that windows appearing at the bottom left corner on darwin is because of https://github.com/andlabs/libui/blob/master/darwin/window.m#L364

The first two parameters `0, 0` are the coordinates for the initial window position.

According to http://www.cocoabuilder.com/archive/cocoa/316913-window-cascading.html, it might be possible to use something like this to cascade windows:

```objectivec
static NSPoint cascadeLoc = {0, 0};
cascadeLoc = [window cascadeTopLeftFromPoint:cascadeLoc];
```

andlabs:
Yes, I know that. Cascading doesn't fix the problem of needing to specify a reference point at all. What do OS X programs created in Xcode use to intelligently decide where on screen they show up?

PR #190, opened 2016-08-08
add #uiOpenFolder

ddo:
tested on windows mac ubuntu all 64 bit

### DIFF
--- a/darwin/stddialogs.m
+++ b/darwin/stddialogs.m
@@ -46,6 +46,20 @@ char *uiOpenFile(uiWindow *parent)
 	return runSavePanel(windowWindow(parent), o);
 }
 
+char *uiOpenFolder(uiWindow *parent)
+{
+	NSOpenPanel *o;
+
+	o = [NSOpenPanel openPanel];
+	[o setCanChooseFiles:NO];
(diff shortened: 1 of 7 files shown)


ddo:
working on next PR: ``#uiSaveFile`` to add ``filename`` as a parameter. so we can set a default file name when saving file.

something like:

```c
uiSaveFile(uiWindow *parent, const char *filename);
```

what do you think @andlabs

andlabs:
The dialog system is frozen right now because of highly incompatible modality requirements across OSs. I'm not sure how I'll fix these things, but I'm not going to be adding or changing the standard dialogs until then.

Not sure whether the parameters to uiSaveFile() should be passed directly or through a structure; it depends on how many extra parameters there will be..

ddo:
yea since open/save dialog has a lot of options as file types, exist path or not, etc... an option struct that describes in common options between OSs would be ok

armand1m:
Why it was never merged? It is broken?

armand1m:
Ok, how can we enable libui since you're not liking it yet? Can we list some tasks that we need to get done for this feature to be implemented?

PR #192, opened 2016-08-11
Fix "cannot find -ldl" linking error with FreeBSD

neelchauhan:
I love the concept of libui, but there will always be people like me who prefer FreeBSD over Linux (or Windows/Mac), so I have made a patch to fix the linking error on FreeBSD. This has been tested working.

Working libui on FreeBSD


andlabs:
I did find this error in FreeBSD back when I used Makefiles; I was hoping cmake would take care of that for me. I guess not :|

And for what it's worth, the concept of libui *includes* FreeBSD, not excludes. I haven't been able to test on the other BSDs since the switch to cmake, but I plan on making binary versions available for [every desktop target supported by Go](https://github.com/golang/go/blob/27c5dcffef590ac9dbc31f3d513f1b7d058907f6/src/cmd/dist/build.go#L1079) (or at least every one I can actually test myself). The only problem is setting up and maintaining a working VM for those systems, especially after switching back to OS X after a laptop change. (Solaris is the biggest problem of all, it seems.)

andlabs:
Also out of curiosity, did you check if the problem happens in both static and shared builds? And does this fix break either static or shared builds? Just to make sure this patch still works in all configurations.

andlabs:
Thanks for the fix! Out of curiosity, what does `$(CMAKE_DL_LIBS)` expand to on FreeBSD?

neelchauhan:
Actually, nothing on FreeBSD, as FreeBSD automatically includes libdl in libc, but I decided to include `${CMAKE_DL_LIBS}` to not break Linux.

I don't know much about CMake either (honestly, my personal C projects just use Makefiles).

andlabs:
Right, having `${CMAKE_DL_LIBS}` is the correct way to do that :) Thanks again!

PR #198, opened 2016-08-19
Add an oo binding reference in d

mogud:
thanks.


andlabs:
OO as in object-oriented? What does the other one do instead?

mogud:
OO, yes. The other one just translate c api to d, nothing more.
libuid minimal programm example:
```d
import ui;
void main() {
    App.run(new Window("hello").addOnClosing(_ => App.quit).show);
}
```

andlabs:
All right, thanks!

PR #202, opened 2016-08-26
Fix enter/escape crashes on Windows

lxn:


andlabs:
Good find; I completely forgot how I handled `WM_COMMAND` there x.x Do you mind adding an explanatory comment above that line to explain the code you added?

ocornut:
FYI even though I don't know libui internals, I stumbled on same crash, googled it, found this patch and it fixed it for me. Thanks!

PaluMacil:
@andlabs  Pietro, would you please consider merging this and also running...
```
git checkout alpha3.1
git checkout -b alpha3.1-maint
```
...to create a maintenance branch of the last binary release? Then you could use that branch to make a tagged binary alpha3.1.1 release that you could move over to the ui Go package. I'd normally recommend git [cherry-pick](https://git-scm.com/docs/git-cherry-pick) to move the change over, but in this case, it's just a tweak to line 85: `if (HIWORD(wParam) != 0 || LOWORD(wParam) <= IDCANCEL)` which might be easiest maintained with a manual fix.

This branch would allow you to make the most urgent of bug fixes to the downstream ui Go package that I love so much. 👍 And this feels like an urgent fix. "Release branches" like this are how a lot of larger projects patch LTS releases, and while you don't have the manpower to have an LTS release, this at least provides a single usable package for us.

PaluMacil:
@andlabs Just following up on my July comment above. If you're able to add this to a new maintenance branch (and cherry pick the commit to the master branch) and then move the fixed alpha3.1 over to the Go repo, it would allow you to have a very low-impact, quick win method of getting the Go library working again without having to rebase onto a new release just yet since you're working on those two feature branches.

andlabs:
I'll handle it after work tonight. It'll probably include everything else that's on `master` right now too, for simplicity's sake.

andlabs:
s/tonight/after I get a copyright clearance from my new job, which is probably not happening until tomorrow or even January/g; sorry about that (I'm frustrated too, but that's what timing brings...)

PaluMacil:
Thanks for the update.

wchristian:
This PR is a little badly structured. The merge commit is entirely unnecessary and the first and second commit should be combined into one.

I recommend doing an interactive rebase where the three commits are squashed into one, followed by a force push to this feature branch. Github will do the right thing.

andlabs:
Sorry I couldn't wait for a response to that last comment, but squash and merge seems to do the right thing, so yay!

PaluMacil:
Thanks for the merge! I'm eager to try this out in Go 1.10 over in the ui repo since I don't have C experience, but I remember a couple things need to happen before that (such as the ldflags fix, I think).

andlabs:
The ldflags fix is already in Go itself. Not sure about binary releases yet.

Issue #184, opened 2016-07-23
Blurry window on Windows and HiDPI displays


emersion:
See:
* https://blogs.msdn.microsoft.com/patricka/2010/04/15/why-does-a-high-dpi-setting-make-my-application-look-fuzzy-and-have-clipped-text/
* https://msdn.microsoft.com/en-us/library/dn469266(v=vs.85).aspx#using_an_application_manifest

andlabs:
High DPI is a TODO.

ebraminio:
There are also some other corner case, multi mon dpi, https://bugs.chromium.org/p/chromium/issues/detail?id=426656 but I feel supporting that would be just unnecessary.

ssylvan:
FWIW I still see blurry text on latest (despite the checkin on Sep 3). See screenshots here. Left is default, middle is using compatibility mode to override it so that the application is in charge of scaling, right is using "System (Enhanced)" scaling.
![image](https://user-images.githubusercontent.com/1365415/33348571-bb983270-d44b-11e7-8117-e61be93f8bf6.png)

This is with 125% DPI scaling. Note that the left is blurry, the right is sharp but too small, and the right looks sharp and the right size. So it seems like it would somehow need to tell windows to do the "enhanced" scaling, whatever that means.

andlabs:
The problem is simply calling `SetProcessDPIAware()` isn't enough to be properly DPI aware; I would need to go through and handle DPI changes correctly, and make sure I"m not accidentally doing pixel-specific stuff.

SatoMew:
Here is [the latest official documentation for Windows](https://msdn.microsoft.com/en-us/library/windows/desktop/mt843498.aspx). Notably, libui needs to support per-monitor v1 and v2 properly so at least Windows 8.1 and 10 Creators Update, respectively, plus the latter's [GDI scaling](https://blogs.windows.com/buildingapps/2017/05/19/improving-high-dpi-experience-gdi-based-desktop-apps/) and the [mixed mode](https://msdn.microsoft.com/en-us/library/windows/desktop/mt744321.aspx) introduced by 10 Anniversary Update. This is because users with high PPI screens are equally likely to use these recent Windows versions.

Windows Vista, 7, and 8 are limited to system DPI awareness and the optional exclusivity of using the old Windows XP scaling or also the new one. [This article](http://kynosarges.org/WindowsDpi.html) is very informative, too.

PR #204, opened 2016-09-03
fixed Windows HiDPI support (#184)

andrew-pa:
It seems like this was didn't require much change... perhaps I missed something but I ran the test program and all the controls (after fixing the color selection button) looked fine at a DPI setting of 144dpi (150% scaling)

### DIFF
--- a/windows/colorbutton.cpp
+++ b/windows/colorbutton.cpp
@@ -61,7 +61,9 @@ static BOOL onWM_NOTIFY(uiControl *c, HWND hwnd, NMHDR *nmhdr, LRESULT *lResult)
 	ID2D1SolidColorBrush *brush;
 	uiWindowsSizing sizing;
 	int x, y;
+	float dpi_x, dpi_y;
 	HRESULT hr;
+	D2D1_MATRIX_3X2_F dm;
 
 	if (nmhdr->code != NM_CUSTOMDRAW)
 		return FALSE;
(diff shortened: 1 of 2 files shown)


andlabs:
Wait, what exactly happened with the color button?

andrew-pa:
The color button rendered its color with Direct2D, and as I remember it calculated the rectangle's size in the wrong units, which match on regular DPI but on high DPI but are slightly off. I don't quite remember right now because it's been a while but Direct2D has some kind of DPI independent units (of course based on regular DPI) and the size was calculated using regular display pixels and then directly given to D2D as device-independent units, so I added some code to change the D2D coordinate system to be in display pixel units so that the rectangle rendered properly. Interestingly enough this issue didn't seem to affect any other stuff rendered with Direct2D. Without the code the colored rectangle was clipped on the right and bottom sides of the button because it was rendered at a larger size than it was supposed to. See this msdn article for detail on the units D2D uses: https://msdn.microsoft.com/en-us/library/windows/desktop/dd756649(v=vs.85).aspx

andlabs:
Oh right, I forgot to convert pixels to DIPs there. Good catch, but I forget if I found this independently as well (which means I have a local fix). Will need to check.

andrew-pa:
Ok will this PR get merged?

andlabs:
I'll merge this when I start pushing my local branch again; a few other PRs here are also waiting for that.

andrew-pa:
Sounds good!

andrew-pa:
Bump? No merge?

andlabs:
Not quite yet; I still have to figure out what else is required for things to truly be DPI-aware, such as making uiArea respond to DPI changes, and also conditional code to enable per-monitor DPI awareness on Windows 8 and newer.

andrew-pa:
Ah quite wise. You're very thorough. That's a good thing. Perhaps close this then?

andlabs:
No; leave it as open as a reminder to myself to get to this. Thanks for understanding!

enbandari:
Any news about this?

PR #208, opened 2016-09-15
Add Swift binding

sclukey:
This provides a nice API to libui from Swift (it's OO, just like the Go version and the new D one and most of the other bindings).

Lower level access to libui (the system module package) is provided by [clibui](https://github.com/sclukey/clibui), but most users shouldn't need that if (and since) libui-swift is feature complete. Anyway if anyone does need it they can find it noted in the first line of libui-swift's readme, so there's no need to muddy this list with it.


andlabs:
Thanks!

Issue #2, opened 2015-09-03
request: uiImage


andlabs:
migrated from https://github.com/andlabs/ui/issues/100

Simple image-displaying widget.

Need to figure out scaling, scrolling, etc.

Need to figure out image loading.

uiArea might obviate the need for this, but putting it up anyway.

trbell:
I'd like to second this request.

HalosGhost:
This is actually one of the only features I'd need to have before I could start using this both for my personal projects but also in-production at ``$DAYJOB``.

art4711:
I made a stab at implementing something that can render images into an Area. I'm not making a pull request yet because I haven't looked at Windows support (and building in Windows makes me want to drink). But I'd like to RFC about a proposed interface.

The libui branch is here: https://github.com/art4711/libui/tree/draw-pixmap
The go ui binding is here: https://github.com/art4711/ui/tree/draw-pixmap
And a test program (with an equivalent gtk program) is here: https://github.com/art4711/fract-ui

The reasoning is explained here:  https://github.com/art4711/libui/blob/draw-pixmap/doc/pixmap.md

Please let me know if it looks reasonable and I should dig into Windows support too, or if it's too far off from a good API I shouldn't waste my time.

andlabs:
Your code is very close to being what I would have wanted. The two main differences would be that

* bitmaps would be pre-created objects like paths rather than ad hoc (which will probably make things harder for uiImage, I'm not sure)
* the byte order would be precisely defined regardless of host endianness

I'll probably drop this in alongside Grid and Table after the next packaged release.

art4711:
I suspect pre-creating the objects would actually make more sense for other applications, my use case is a blank canvas of pixel sized pixels. So I just want the library to render things and not get involved otherwise.

I'll look into how much it'd take to do this. The big question is scaling. The way the api is designed things are supposed to be as unaware of pixels as possible. If cairo, quartz and windows can scale for us, that should be easy. If not, this will take a while.

The byte order part. Like I wrote, Cairo expects host-endian only, no options there. Quartz can specify endianness. So any explicit big or little endian would require Cairo rendering to manually convert each uint32 when copying the pixmap. It's significantly slower to do that. Right now, all the conversion needed (flip the image upside down for Quartz) is one memcpy per row of data which is as cheap as it can get.

[...]

PR #209, opened 2016-09-15
Create a uiImage control

sclukey:
This adds a simple control for loading an image from a file and getting and setting its size.

This shouldn't invalidate the work done in #2 by @art4711 since most GUI systems have a pixel buffer and an image component (Gtk has GdkPixbuf -> GtkImage, Windows has HBITMAP -> HWND, Cocoa has NSImage -> NSImageView, Qt has QPixmap -> QLabel, etc.). Perhaps in the future the uiImage control should be backed by a uiPixbuf or something, but for now at very least there should be a simple component for displaying an image.

### DIFF
--- a/common/controlsigs.h
+++ b/common/controlsigs.h
@@ -13,6 +13,7 @@
 #define uiFormSignature 0x466F726D
 #define uiGridSignature 0x47726964
 #define uiGroupSignature 0x47727062
+#define uiImageSignature 0x4723BDC6
 #define uiLabelSignature 0x4C61626C
 #define uiMultilineEntrySignature 0x4D6C6E45
 #define uiProgressBarSignature 0x50426172
--- a/darwin/CMakeLists.txt
+++ b/darwin/CMakeLists.txt
@@ -21,6 +21,7 @@ list(APPEND _LIBUI_SOURCES
 	darwin/form.m
 	darwin/grid.m
 	darwin/group.m
+	darwin/image.m
 	darwin/label.m
 	darwin/main.m
 	darwin/map.m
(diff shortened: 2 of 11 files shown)


andlabs:
I already have a design for image loading in my local repository; it won't work in a way that's compatible with this. Thanks in the meantime, though!

sclukey:
Ok no problem. Perhaps would you mind pushing that in a feature branch or something?

I see what you are saying [here](https://github.com/andlabs/libui/issues/2#issuecomment-228571416), but that only seems to deal with application resources, not external images (which generally only have a single representation regardless of screen resolution). Even if there are multiple resolutions then Cocoa handles that directly with representations and a similar thing could be tacked onto Gtk and Win32. Also I'm not sure what the purpose of the floating point sizes is, it seems to just be a pixel size and a scale factor.

Anyway, I guess I'm just curious what you're up to and I'd be willing to help out with your version if you'd like it.

Swyter:
Same, I have been wanting to use this. This component looks pretty trivial and the implementation is pretty straightforward in the three back-ends, seems like a good fit for the type of simple dialog windows that one would make with libui. I think.

mischnic:
>  Any luck with image support? I know this is 2 years old now

From #318:
> This is dependent on #310, as the table branch has a lot of the groundwork done already. Once that's done, I'll start discussing and planning the API (as I need to go back through my notes and the OS API documentation and also make sure I'm doing high-DPI multi-resource correctly).

gregbenner:
Awesome! great project, will be really exiting to have images. Would be able to do really professional apps with https://proton-native.js.org/#/

andlabs:
As soon as #361 is merged in, and table merged with master some hopefully short time later, I'll have a uiImage type that you can load bitmaps into, and we can talk about a control from there.

mischnic:
So this would need to be updated to use the existing `uiImage`, although that one is really meant for icons with multiple resolutions?

mischnic:
Also, a test/example is missing.

PR #222, opened 2016-10-29
uiBoxNumControls

krakjoe:
Determining the number of controls is important for avoiding errors.

I only checked the unix one, and am very unsure about the objc one.


andlabs:
Why did you delete this?

krakjoe:
The darwin one was broken, and needed access to private fields, I didn't know that when I suggested it. 

I guess it's not really acceptable to change that ?

krakjoe:
I'll leave you to decide what to do with it ;)

andlabs:
You'd just add a method to the uiBox type for that =P I'll take care of it if you want.

krakjoe:
This is now complete, and working on darwin.

Issue #193, opened 2016-08-11
Clipboard support


Impyy:
GUI applications often require access to the clipboard. Is this something you'd like to have implemented in libui or is that out of the scope of this project? On Linux, for example, we already link against GTK so we might as well make use of the clipboard API that it has. If it's not considered out of scope, I'll try to implement clipboard support and submit a PR. I'll need some help with the OSX part though as I don't own a mac and I don't speak Objective C.

andlabs:
Planned for the future. I'm not sure what data types I would need to handle or how to handle them.

Impyy:
Like I said, I'm willing to help out. I could start with just text support (which should be simple enough) and the API can be extended to support other data types later. How does that sound?

mardukbp:
FYI, [libclipboard](https://github.com/jtanx/libclipboard) is a cross-platform clipboard library implemented in C.

PR #224, opened 2016-10-30
[WIP] Clipboard support

Impyy:
After not getting a response to my question on #193 I decided to go ahead and start working on this feature. As you can tell, this is still a work in progress but I'm opening this now so anyone can comment on the API if they want to. I only plan on implementing text support for now.

Progress:
- [x] Unix
- [ ] Windows
- [ ] OSX

Windows support is in the works but I'll need some help with OSX. I'm also not very experienced with C so please do yell at me if I'm doing something weird.


andlabs:
I'm not sure how I'm going to work with non-text types.

I wonder why GTK+'s API requires an event loop and callback...

Impyy:
@andlabs 
>I wonder why GTK+'s API requires an event loop and callback...

It's because of the way clipboards work on X. If the clipboard content is in the same process it can be retrieved directly whereas if the content is in a different process we have to wait for the process to respond with the content.

I just noticed that there's gtk_clipboard_wait_for_text. It runs GTK's main loop and returns once the clipboard content is received. We may be able to use this instead.

Impyy:
Ok. I've changed it to use gtk_clipboard_wait_for_text.

andlabs:
Don't change it just yet. I need to see if the Windows versions also do (under some circumstances, such as `OleGetClipboard()`?), and which APIs to use.

As for X11 not providing a true synchronous version, even though I understand *why*,

lame

Impyy:
Too late, lol. I'll revert the change if you find out if we can't avoid an asynchronous API.

Impyy:
I've added support for windows. It looks like the API isn't asynchronous, luckily. Let me know if you want any changes.

andlabs:
My issue isn't if the API is asynchronous, but rather if the API pumps messages while waiting, if it needs to wait. I'll have to ask about that.

Issue #116, opened 2016-06-08
Add install rule


HalosGhost:
At the moment, the `Makefile` generated by `cmake` does not include an `install` rule.

To make packagers' lives a lot simpler, it would be great if this could be added.

It seems like it should be as simple as making a few calls to `install()` in the `CMakeLists.txt` files to put everything where it needs to go.

andlabs:
`libui` itself needs to be installed, as well as `ui.h` and `ui_(current OS).h`. Do the other OS files need to be installed as well? And how would I specify the headers?

HalosGhost:
The library needs to be installed (e.g., ``libui.so.0`` and symlinks if applicable—e.g., ``libui.so``), and the headers need to be installed (that are relevant to the system).

At some point (perhaps this should be a separate issue or a PR), but having a pkg-config file would be helpful for people.

As for how to "specify the headers", I suppose I am not sure what you mean.

andlabs:
I meant what cmake syntax do I use to do this

HalosGhost:
Looks like the [``cmake`` docs](https://cmake.org/cmake/help/v3.0/command/install.html) are the best place to check.

I am not personally that familiar with cmake; my understanding of what is necessary to complete this comes from inferences drawn by looking at other projects that have done the same thing.

mpiannucci:
CMake should be installed with a Config.cmake file that allows one to find and include it with `find_package(libui)`. There is a tutorial for how to do this here https://cmake.org/Wiki/CMake/Tutorials/How_to_create_a_ProjectConfig.cmake_file

SirWindfield:
That would be good to have. I use xgo to cross compile libui for each OS and the utility assumes that the libraries are properly installed via `make install`. Sadly, libui doesn't provide the target for `install`. Would be nice to have it.

mpiannucci:
This is a pretty simple task. I can try to tackle it in the the coming week or so.

andlabs:
Hi I forgot this wasn't solved yet. Why did I wait on it?

andlabs:
Okay, so will this be sufficient?

```cmake
install(TARGET ${_LIBUINAME}
    INCLUDES DESTINATION include
    LIBRARY DESTINATION lib
    ARCHIVE DESTINATION lib)
```

? Will the include stuff only work if I have a `FRAMEWORK` instead of a `LIBRARY`? It should already include only what we need.

[...]

PR #232, opened 2016-11-04
CMake install

krakjoe:
As requested by #116

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -217,3 +217,15 @@ macro(_add_exec _name)
 endmacro()
 add_subdirectory("test")
 add_subdirectory("examples")
+
+install(TARGETS ${_LIBUINAME}
+   LIBRARY DESTINATION lib
+   ARCHIVE DESTINATION lib)
+install(FILES ui.h uitable.h DESTINATION include)
+if (APPLE)
+	install(FILES ui_darwin.h DESTINATION include)
+elseif (WIN32)
+	install(FILES ui_windows.h DESTINATION include)
+else()
+	install(FILES ui_unix.h DESTINATION include)
+endif()


philip:
@krakjoe  It appears `uitable.h` should be removed from this patch as it was moved to a `table` branch.

CMakeLists.txt
```diff
@@ -217,3 +217,15 @@ macro(_add_exec _name)
 endmacro()
 add_subdirectory("test")
 add_subdirectory("examples")
+
+install(TARGETS ${_LIBUINAME}
+   LIBRARY DESTINATION lib
+   ARCHIVE DESTINATION lib)
+install(FILES ui.h uitable.h DESTINATION include)
```
rwols: Do we still need `uitable.h`?

CMakeLists.txt
```diff
@@ -217,3 +217,15 @@ macro(_add_exec _name)
...
+if (APPLE)
+	install(FILES ui_darwin.h DESTINATION include)
+elseif (WIN32)
+	install(FILES ui_windows.h DESTINATION include)
+else()
+	install(FILES ui_unix.h DESTINATION include)
+endif()
```
rwols: You can just do `install(FILES ui.h ui_${_OSNAME}.h DESTINATION include)`, no?

In fact, wouldn't it be even better if we did `set_target_properties(${_LIBUINAME} PROPERTIES PUBLIC_HEADER ui.h ui_${_OSNAME}.h)`, and then `install(TARGETS ${_LIBUINAME} LIBRARY DESTINATION lib RUNTIME DESTINATION lib PUBLIC_HEADER DESTINATION include)`?

andlabs:
No longer relevant given the upcoming (either tomorrow or Tuesday) switch to Meson. In fact, I've also set up the install rules in this new environment, so you finally get your install rules. Thanks anyway!

PR #238, opened 2016-11-30
Move uiImage declaration to ui_unix.h

franko:
Needed otherwise alloc.c does not compile on unix.


franko:
Libui fails to compile on ubuntu. I have tried to move the declaration of uiImage into the appropriate header file to fix the problem.

andlabs:
Moving the declaration wasn't the right solution; I fixed it with a cleaner hack for now. The problem was I didn't split table into its own branch fully; sorry about that!

PR #240, opened 2016-12-01
Update uipriv_darwin.h

lailton:
Fix unknown type name 'uiImage'


andlabs:
Can you add a comment `// TODO remove this after tables go back in` as well?

lailton:
Ok

andlabs:
All right, thanks!

PR #247, opened 2017-01-11
Resource copying on windows for VC++ 32/64 mingw and msys

ptriller:
There are still some issues with resource copying on Windows. Not all environments support wildcard expansion. I found a solution with the builtin variable CMAKE_RC_OUTPUT_EXTENSION and tested it with all environments I could easily do:
 - Visual C++ 2015 x64 Native Build Tools DLL and static
 - Visual C++ 2015 x86 Native Build Tools DLL and static
 - MSYS2/mingw64 in bash: static
 - MSYS2/mingw64 in cmd.exe: static
 - JetBrains CLion with mingw64 toolchain

And all configurations are able to compile and start the "controlgallery" example.

### DIFF
--- a/windows/CMakeLists.txt
+++ b/windows/CMakeLists.txt
@@ -72,7 +72,7 @@ macro(_handle_static)
 	add_custom_command(
 		TARGET libui POST_BUILD
 		COMMAND
-			${CMAKE_COMMAND} -E copy $<TARGET_PROPERTY:libui,BINARY_DIR>/CMakeFiles/libui.dir/windows/resources.rc.* ${_LIBUI_STATIC_RES}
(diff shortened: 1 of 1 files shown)


andlabs:
Just rediscovered this now. Leaving a note to myself for when I'm ready to merge PRs: [`CMAKE_RC_OUTPUT_EXTENSION` support for MinGW was introduced with cmake 2.8.4](https://github.com/Kitware/CMake/commit/b2f308c8f9b4ded77f0693b39c0288e3bbb00977) ([though the variable itself is older, dating back to at least 2.4.0](https://github.com/Kitware/CMake/commit/692ba48c4e5762b370f2999e902b8bd677c77161)). It's a shame not all of these special variables are fully documented :/

andlabs:
I've decided to drop copying the resource file anyway; see #308 (which this is merged with). Thanks in the meantime!

PR #249, opened 2017-01-22
Fix release build on MSVC.

l0calh05t:
The /RTC* flags conflict with all optimizations, therefore they should only be added when building the Debug configuration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,7 @@ if(MSVC)
 	set(_COMMON_CFLAGS
 		/W4 /wd4100
 		/bigobj /nologo
-		/RTC1 /RTCs /RTCu
+		$<$<CONFIG:Debug>:/RTC1 /RTCs /RTCu>
 		/EHsc
 	)
 


andlabs:
I thought I asked this elsewhere, but I guess I forgot to: where does that information come from? I still need to figure out what Visual Studio does differently between the two configurations by default in terms of compiler and linker options...

andlabs:
Actually I see now I asked this in #261, but only partially. I wonder if we can merge these two PRs somehow...

slahn:
To try and answer you question, this is documented for the /RTC*-flags at least: https://docs.microsoft.com/nb-no/cpp/build/reference/rtc-run-time-error-checks.

I'm really no expert with CMake or Visual C++, but I don't think the Visual Studio defaults matter in this case, what matters is that CMake writes to the project/msbuild files.

Here, your CMake files are telling the compiler to use two incompatible options at the same time, because you add the RTC-flags to the CMake defaults for release builds.

The Cmake defaults are specified in `/usr/share/cmake-$version/Modules/Platform/Windows-MSVC.cmake`, or `$cmakeInstallDir\share\cmake-$version\Modules\Platform\Windows-MSVC.cmake` on Windows:
`string(APPEND CMAKE_${lang}_FLAGS_RELEASE_INIT " /MD /O2 /Ob2 /DNDEBUG")`

l0calh05t:
The other pull request only works for single-configuration generators, so please don’t use it.

mischnic:
> is the other PR's use of Debug vs DEBUG correct in general? This is one part where cmake's case-insensitivty confuses me.

https://cmake.org/pipermail/cmake/2013-June/055177.html:
> I see that this CMake variable is not case-sensitive at it and defaults 
to DEBUG.

<br>
In the documentation, it's "Debug": https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html

---

> does the minimum version of cmake in CMakeLists.txt match that needed by the $<> directive in this one?

These are [generator expressions](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html). 
The very expression in this PR (`$<$<CONFIG:Debug>:DEBUG_MODE>`) is described in the cmake [docs for version 3.0](https://cmake.org/cmake/help/v3.0/manual/cmake-generator-expressions.7.html#logical-expressions) and 3.1. So we should be okay with `3.1.0` as `cmake_minimum_required `.

andlabs:
All right; thanks. I'll merge this one and decide what to do with the other one.

PR #254, opened 2017-02-12
Add .NET Core bindings to readme

nopara73:


andlabs:
Out of curiosity, what's the difference between .NET Framework and .NET Core?

nopara73:
.NET Core is cross platform.

andlabs:
Thanks!

PR #261, opened 2017-04-01
Fix MSVC Release build.

zentner-kyle:
Previously, the CMakeLists.txt unconditionally used the /RTC1 /RTCs /RTCu
flags to enable run time checks. However, these flags are not compatible
with optimizations. If one attempted to build libui using
CMAKE_BUILD_TYPE=Release, this would cause the build to fail, since the
MSVC driver uses /02 in Release mode.

This commit only enables those flags in Debug mode, so that they are still
useful for debugging, and Release builds can still be made. Being able to
build in Release mode is useful, because by default other projects
depending on libui may need to build it in Release mode.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,15 +71,22 @@ if(MSVC)
 	# TODO /sdl turns C4996 into an ERROR
 	# don't use /analyze; that requires us to write annotations everywhere
 	# TODO undecided flags from qo?
-	# /RTCc is not supplied because it's discouraged as of VS2015; see https://www.reddit.com/r/cpp/comments/46mhne/rtcc_rejects_conformant_code_with_visual_c_2015/d06auq5
 	# /EHsc is to shut the compiler up in some cases
 	# TODO make /EHsc C++-only
 	set(_COMMON_CFLAGS
 		/W4 /wd4100
 		/bigobj /nologo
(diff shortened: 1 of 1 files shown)


andlabs:
Is that specific `if()` condition correct? My intuition tells me it shouldn't be, but I'm not a CMake expert...

zentner-kyle:
Sorry for the slow response.

It was correct, but only if the CMAKE_BUILD_TYPE wasn't explicitly set and the logic at the top of the CMakeLists.txt file set it to "DEBUG". Usually, the string "Debug" would be used, so this was somewhat confusing. I've added a new commit which makes this more consistent.

I'm also not a CMake expert, but I have tested that the `if()` condition triggers in the correct case (CMAKE_BUILD_TYPE is "Debug"), and doesn't trigger when CMAKE_BUILD_TYPE is "Release".

Intuitively, what I wrote before probably didn't look like it would be correct because CMAKE_BUILD_TYPE is a variable, which gets expanded (usually to "DEBUG"), but DEBUG is a bare word which just becomes the string "DEBUG". CMake is just weird like that, I guess.

I could change `CMAKE_BUILD_TYPE` to `${CMAKE_BUILD_TYPE}`, if that makes the code more clear to you.

andlabs:
See discussion on #249 as I need to resolve the differences between this one and that one before taking action.

Issue #223, opened 2016-10-29
compile failed on windows by mingw-w64 


walkblank:
Here is the failed information:

Error copying file "/CMakeFiles/libui.dir/windows/resources.rc.*" to "F:/libui-master/libui-master/build/out/libui.r

CMakeFiles\libui.dir\build.make:1678: recipe for target out/libui.a' failed make[2]: **\* [out/libui.a] Error 1 make[2]: **\* Deleting fileout/libui.a'
CMakeFiles\Makefile2:66: recipe for target CMakeFiles/libui.dir/all' failed make[1]: **\* [CMakeFiles/libui.dir/all] Error 2 makefile:82: recipe for targetall' failed
make: **\* [all] Error 2

snuk182:
Confirming issue as well. Looks like cmake for windows does not accept copying multiple sources to the single target (resources.rc.* => libui.rs).

snuk182:
Although this PR fits better: https://github.com/andlabs/libui/pull/247

andlabs:
Merged into #308.

PR #263, opened 2017-04-12
Fix for Win64 + GNU + staticlib build

snuk182:
https://github.com/andlabs/libui/issues/223


snuk182:
https://github.com/andlabs/libui/pull/247 might be a broader solution.

PR #265, opened 2017-04-21
add opengl draw capability to area

stonewell:
add opengl draw capability to area


andlabs:
GtkGLArea is too new at present to use.

Could you describe what you added in more detail?

@pcwalton, how does this compare with your code?

PR #269, opened 2017-06-23
Set stdlib to clang's libc++ to unbreak the branch build on macOS

ebraminio:
This is to unbreak my build on macOS Sierra, this seems is a solution https://stackoverflow.com/a/11972743

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -27,6 +27,11 @@ _add_example(cpp-multithread
 	cpp-multithread/main.cpp
 	${_EXAMPLE_RESOURCES_RC}
 )
+if(APPLE)
+	# forcing libc++ headers and runtime on Apple in order to have access to C++11 stuffs
+	target_compile_options(cpp-multithread PUBLIC -stdlib=libc++)
+	target_link_libraries(cpp-multithread -stdlib=libc++)
+endif()
 if(NOT WIN32)
 	target_link_libraries(cpp-multithread pthread)
 endif()


andlabs:
As I mentioned on Telegram, I didn't have this build problem on El Capitan, but I do now on Sierra. I wonder what changed on 10.11...

Also the only C++ on OS X right now is the `cpp-multithread` example, so I wonder if it should be made specific to that project (see `examples/CMakeLists.txt`). Or would that break other C++ projects that want to link to libui (either statically or dynamically)?

ebraminio:
> Also the only C++ on OS X right now is the cpp-multithread example, so I wonder if it should be made specific to that project (see examples/CMakeLists.txt). Or would that break other C++ projects that want to link to libui (either statically or dynamically)?

Yes, I had the issue `cpp-multithread`, whatever you wish is OK to me

ebraminio:
with this patch, unlike the current patch,

```diff
diff --git a/examples/CMakeLists.txt b/examples/CMakeLists.txt
index f2289d2..d4ade09 100644
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -30,6 +30,10 @@ _add_example(cpp-multithread
 if(NOT WIN32)
        target_link_libraries(cpp-multithread pthread)
 endif()
+# forcing libc++ use on Apple in order to get access to newer stuffs
+if (APPLE)
+       target_compile_options(cpp-multithread PUBLIC -stdlib=libc++)
+endif ()
 
 _add_example(drawtext
        drawtext/attributes.c
```

I am getting this error:

```
clang: warning: libstdc++ is deprecated; move to libc++ with a minimum deployment target of OS X 10.9
Undefined symbols for architecture x86_64:
```

I am not sure why but it is interesting...

andlabs:
I hope that message doesn't show up regardless of the deployment target...

ebraminio:
Ops, obviously I should have put the flag on link time also, fixed.

PTAL

ebraminio:
> I hope that message doesn't show up regardless of the deployment target...

Fortunately not but lots of:

```
...area.m:164:7: warning: expression is not an
      integer constant expression; folding it to a constant is a GNU extension
      [-Wgnu-folding-constant]
        case NSOtherMouseDown:
             ^~~~~~~~~~~~~~~~
```

ebraminio:
Applied on https://github.com/andlabs/libui/commit/242620ff0f111898e61f84b1dc758cc0be67c28d

andlabs:
Oh, I forgot this was there and just did the fix manually. Sorry!

PR #270, opened 2017-06-27
Better default position for window on macOS

ebraminio:
Instead default putting windows on bottom-right, this sets default position of windows to a better place, similar to jdk's [nativeSetNSWindowLocationByPlatform](https://github.com/netroby/jdk9-dev/blob/44ab3e51cec57c11f896fe8f5df45a7bf9b4a07f/jdk/src/java.desktop/macosx/native/libawt_lwawt/awt/AWTWindow.m#L1163-L1186).

### DIFF
--- a/darwin/window.m
+++ b/darwin/window.m
@@ -367,6 +367,8 @@ static void defaultOnPositionContentSizeChanged(uiWindow *w, void *data)
 	// do nothing
 }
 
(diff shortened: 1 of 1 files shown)


ebraminio:
This should be reopened on the master branch

ebraminio:
Closed in favor of #271

PR #271, opened 2017-06-27
Better default position for window on macOS

ebraminio:
Instead default putting windows on bottom-right, this sets default position of windows to a better place, similar to jdk's [nativeSetNSWindowLocationByPlatform](https://github.com/netroby/jdk9-dev/blob/44ab3e51cec57c11f896fe8f5df45a7bf9b4a07f/jdk/src/java.desktop/macosx/native/libawt_lwawt/awt/AWTWindow.m#L1163-L1186).

There is a question however, should we apply that on [onVisible](https://github.com/netroby/jdk9-dev/blob/44ab3e51cec57c11f896fe8f5df45a7bf9b4a07f/jdk/src/java.desktop/macosx/classes/sun/lwawt/macosx/CPlatformWindow.java#L551-L561) or the current approach on applying it on newWindow is OK? I guess for the scope of libui the current approach is fine.

### DIFF
--- a/darwin/window.m
+++ b/darwin/window.m
@@ -367,6 +367,8 @@ static void defaultOnPositionContentSizeChanged(uiWindow *w, void *data)
 	// do nothing
 }
 
+static NSPoint lastTopLeftPoint;
+
 uiWindow *uiNewWindow(const char *title, int width, int height, int hasMenubar)
 {
 	uiWindow *w;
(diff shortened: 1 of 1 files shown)


ebraminio:
An approach very similar to JDK I guess would be having setPositionByPlatform flag somewhere and apply that only in case of position is not set previously by the library user, as that probably contains some design change (and maybe doing something similar on Windows?) feel free to close this PR and apply it on that way, I just wanted to not forget this.

ebraminio:
There is also another solution, putting new window on [center](https://github.com/glfw/glfw/blob/4dee58d480dc034d0cc188c7a7f2cc9328f5c1d1/src/cocoa_window.m#L1032) that glfw is doing that which as the different scope of these two projects, a different approach makes sense.

PR #275, opened 2017-07-20
[PoC] macOS vibrancy

ebraminio:
just a proof of concept for macOS vibrancy, for now. welcomes your comments also but if not interested, feel free to just close it :)

### DIFF
--- a/darwin/window.m
+++ b/darwin/window.m
@@ -1,5 +1,6 @@
 // 15 august 2015
 #import "uipriv_darwin.h"
+#import <string.h>
 
 #define defaultStyleMask (NSTitledWindowMask | NSClosableWindowMask | NSMiniaturizableWindowMask | NSResizableWindowMask)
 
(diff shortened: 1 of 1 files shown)


mischnic:
It would be really cool if this was a property of a container (maybe additionally to the window):

bildschirmfoto 2018-11-28 um 12 34 25

Do GTK and Windows support something similar?

andlabs:
There are a number of things there that I do want to have in libui.

The sidebar itself is definitely possible in all toolkits, but the glass effect is a bit more complicated. With GTK+ it's up to the theme how it chooses to render stuff. With Windows there's DWM blur-behind for the client area, but that's not quite what that is; there's also UWP blur-behind, but for whatever reason that's not available to Win32 programs unless you're already using UWP anyway.

That particular window layout is something I want to do as well. But these are high-level ideas for libui... ideas I haven't fully figured out yet. (On the other hand, I have started exploring headerbar-style controls on Windows that look like the one in Explorer.)

PR #291, opened 2018-01-25
Crash on pressing enter or escape

Vishal1707:


andlabs:
Submitting someone else's pull request as your own isn't going to make me more inclined to take it, especially since it also includes extra work that I don't know anything about.

Vishal1707:
Apologies @andlabs n @lxn, it was not intentional.

andlabs:
It's fine; just note that you should see if someone has already made an open PR for a fix before making one yourself, and watch out for any extra commits that leak in.

andlabs:
The main fix is already in master.

The other commit this includes now has separate discussion in https://github.com/Araq/libui/commit/0cc893df6f67d6c2f9e2ce3e6c24b66a1efaaf2a, since it's incorrect anyway.

Sorry again for yelling.

Issue #295, opened 2018-02-18
Getting value from uiDateTimePicker?


LeoTindall:
How would I go about getting the time and date value from a uiDateTimePicker?

andlabs:
I haven't written that part yet, sorry. I do need to get back to rounding out the regular controls.

LeoTindall:
No worries! Just wanted to make sure I wasn't missing something.

I appreciate all the hard work you've done in this project so far. Thanks!

andlabs:
Will keep open as a reminder anyway.

cody271:
Welcome back, @andlabs! Proper handling of date and time can be really tricky, but I can work on this issue. Can we use `struct tm` from the C standard library, or do we need a `uiDateTime` struct?

What do you think of this for the `ui.h` API?
```
#include <time.h>

_UI_EXTERN void uiDateTimePickerTime(uiDateTimePicker *d, struct tm *time);
_UI_EXTERN void uiDateTimePickerSetTime(uiDateTimePicker *d, struct tm *time);
_UI_EXTERN void uiDateTimePickerOnChanged(uiDateTimePicker *d, void (*f)(uiDateTimePicker *, void *), void *data);
```

andlabs:
I don't know, but I do know uiDateTimePicker won't handle any actual time stuff, and should just return the time as the user has entered it. Programs would be responsible for properly doing time zone calculations.

cody271:
Cool I will start with `struct tm` then. Does that API otherwise look acceptable to you before I starting implementing?

andlabs:
It does.

I'm not sure if `<time.h>` should be included or `struct tm;` should be predeclared.

Also when filling a `struct tm`:

- `tm_wday` and `tm_yday` should be ignored; documentation should say their values should be undefined
- `tm_isdst` should be negative to signify "unknown"; this should be documented

When reading from a `struct tm` it *might* be safe to just ignore those three fields; IDK.

I'm not particularly thrilled that there is no UTC `mktime()`, nor one that allows you to specify an arbitrary time zone, but whatever.

andlabs:
Replaced with #297.

PR #297, opened 2018-02-22
Improve uiDateTimePicker

cody271:
uiDateTimePickerTime() APIs and example (#295)


darwin/datetimepicker.m
```diff
@@ -4,17 +4,90 @@
...
+	return self;
+}
+
+- (void)dealloc
+{
+	mapDestroy(self->pickers);
+	[super dealloc];
+}
+
+- (void)datePickerCell:(NSDatePickerCell *)aDatePickerCell
```
andlabs: NSDatePicker is a NSControl, so target-action *should* work and you won't need to fiddle with `NSTimer` or delegates. If it doesn't, could you add a comment explaining that (and consider using `dispatch_async()` instead of `NSTimer`)?

darwin/datetimepicker.m
```diff
@@ -40,3 +120,36 @@
...
+
+void uiDateTimePickerTime(uiDateTimePicker *d, struct tm *time)
+{
+	time_t t;
+	struct tm tmbuf;
+	NSDate *date;
+	NSTimeInterval interval;
+
+	date = [d->dp dateValue];
+	interval = [[NSTimeZone systemTimeZone] secondsFromGMTForDate:date];
```
andlabs: What do this line and the next one do? I'm not sure if I'm comfortable using NSTimeInterval since that's really a double and (from what I can gather) `time_t` is not; what about `NSDateComponents` (with `[d->dp calendar]` as the source calendar to extract the components with)? I forget if `NSDate` has something more precise than `NSTimeInterval` as an option (maybe Core Foundation does via toll-free bridging?)...

darwin/datetimepicker.m
```diff
@@ -40,3 +120,36 @@
 {
 	return finishNewDateTimePicker(NSHourMinuteSecondDatePickerElementFlag);
 }
+
+void uiDateTimePickerTime(uiDateTimePicker *d, struct tm *time)
```
andlabs: Could these methods be moved above the constructor, like in the other files?

unix/datetimepicker.c
```diff
@@ -111,8 +122,11 @@ static void setLabel(dateTimePickerWidget *d)
 
 static void dateTimeChanged(dateTimePickerWidget *d)
 {
+	uiDateTimePicker *c;
```
andlabs: Honestly, I would make this emit an actual signal and separate the widget implementation from the uiDateTimePicker implementation. But if you can't figure out how to do that, don't worry (I'll probably take care of that instead).

windows/datetimepicker.cpp
```diff
@@ -189,3 +189,15 @@ uiDateTimePicker *uiNewTimePicker(void)
 {
 	return finishNewDateTimePicker(DTS_TIMEFORMAT);
 }
+
+void uiDateTimePickerTime(uiDateTimePicker *d, struct tm *time)
```
andlabs: For this, you will have to populate all the fields of `struct tm` manually from a `SYSTEMTIME` structure. I can take care of it if you'd prefer.

darwin/datetimepicker.m
cody271: `- (void)datePickerCell:` allows you to modify or even reject the event via `*proposedDateValue`, so `dateValue` isn't valid until the method returns. The timer moves the actual event firing to the next run of the event loop, and it could totally use `dispatch_async()` instead of `NSTimer` if it makes a difference.

darwin/datetimepicker.m
cody271: Will do. `libui` constructors go at the end of the file. I didn't pickup on that convention until just now.

windows/datetimepicker.cpp
cody271: I wanted to get your feedback before working on the Windows implementation. Will do it next after fixing the other issues you've brought up here.

unix/datetimepicker.c
cody271: Will emit a GTK+ signal like a proper widget then. Agreed that is better separation of concerns.

darwin/datetimepicker.m
cody271: Honestly have no idea why the `NSTimeZone` bit is needed! It makes the example program work. Without it, the dates returned are totally wrong. (`NSDatePicker` is weird!)

Agreed `double` to `time_t` conversation looks really bad. There is no portable size guarantee for `time_t`. 
Though I think `NSTimeInterval` can be truncated to an integer type okay since we only need resolution to the second.

I'll have to research more.. and definitely could use a bit of help on these two.

darwin/datetimepicker.m
cody271: `- (void)datePickerCell:` allows you to modify or even reject the event via `*proposedDateValue`, so `dateValue` isn't valid until the method returns. The timer moves the actual event firing to the next run of the event loop, and it could totally use `dispatch_async()` instead of `NSTimer` if it makes a difference.

darwin/datetimepicker.m
cody271: Honestly have no idea why the `NSTimeZone` bit is needed! It makes the example program work. Without it, the dates returned are totally wrong. (`NSDatePicker` is weird!)

Agreed `double` to `time_t` conversation looks really bad. There is no portable size guarantee for `time_t`. Though I think `NSTimeInterval` can be truncated to an integer type okay since we only need resolution to the second.

I'll have to research more.. and definitely could use a bit of help on these two.

darwin/datetimepicker.m
cody271: Will do. libui constructors go at the end of the file. I didn't pickup on that convention until just now.

unix/datetimepicker.c
cody271: Will emit a GTK+ signal like a proper widget then. Agreed that is better separation of concerns.

windows/datetimepicker.cpp
cody271: I wanted to get your feedback before working on the Windows implementation. Will do it next after fixing the other issues you've brought up here.

darwin/datetimepicker.m
cody271: Fixed the `NSTimeZone` nonsense!

`NSTimeInterval` to `time_t` maybe still an issue.

darwin/datetimepicker.m
cody271: Fixed.

windows/datetimepicker.cpp
cody271: Windows seems to need `tm_wday` set correctly for some reason.

cody271:
As of aac0f86, `datetime` example program now works on all three platforms!

windows/datetimepicker.cpp
```diff
@@ -131,6 +134,72 @@ static void uiDateTimePickerMinimumSize(uiWindowsControl *c, int *width, int *he
...
+	if (nmhdr->code != DTN_DATETIMECHANGE)
+		return FALSE;
+	(*(d->onChanged))(d, d->onChangedData);
+	*lResult = 0;
+	return TRUE;
+}
+
+static void fromSystemTime(LPSYSTEMTIME systime, struct tm *time)
+{
+	memset(time, 0, sizeof(struct tm));
```
andlabs: For consistency, use `ZeroMemory()` on Windows.

windows/datetimepicker.cpp
```diff
@@ -131,6 +134,72 @@ static void uiDateTimePickerMinimumSize(uiWindowsControl *c, int *width, int *he
...
+static void defaultOnChanged(uiDateTimePicker *d, void *data)
+{
+	// do nothing
+}
+
+void uiDateTimePickerTime(uiDateTimePicker *d, struct tm *time)
+{
+	SYSTEMTIME systime;
+
+	if (DateTime_GetSystemtime(d->hwnd, &systime) != GDT_VALID)
```
andlabs: For consistency, use the `SendMessageW()` messages directly, instead of the wrapper macros.

windows/datetimepicker.cpp
```diff
@@ -131,6 +134,72 @@ static void uiDateTimePickerMinimumSize(uiWindowsControl *c, int *width, int *he
...
+		implbug("DTM_GETSYSTEMTIME message failed");
```
andlabs: Use `logLastError()` here.

cody271:
Fixed the Windows implementation consistency issues in 3f228c0, any other problems here left?

PaluMacil:
Thanks for the work on this, @cody271 I really appreciate it!

I'm excited to see it merged, but I'm probably going to take a while to wade through the code since I write Go rather than C.

cody271:
@PaluMacil It would be great if you could build & run the new example program [`datetime`](https://github.com/cody271/libui/blob/3f228c0d3541b941ea4cec09b83704777e87b5eb/examples/datetime/main.c).

andlabs:

darwin/datetimepicker.m
andlabs: Using `dispatch_async()` looks more correct, and comes with the guarantee of being run when next possible, rather than being subject to any delays introduced by NSTimer.

unix/datetimepicker.c
```diff
@@ -532,6 +541,68 @@ static void dateTimePickerWidget_class_init(dateTimePickerWidgetClass *class)
 {
 	G_OBJECT_CLASS(class)->dispose = dateTimePickerWidget_dispose;
 	G_OBJECT_CLASS(class)->finalize = dateTimePickerWidget_finalize;
+
+	changedSignal = g_signal_new("changed",
+		G_TYPE_FROM_CLASS(class),
+		G_SIGNAL_RUN_LAST,
+		0,
+		NULL, NULL, NULL,
+		G_TYPE_NONE, 0);
```
andlabs: I'd put the last argument on its own line since it denotes the number of arguments. (I wonder if we even need a signal at all...)

unix/datetimepicker.c
```diff
@@ -532,6 +541,68 @@ static void dateTimePickerWidget_class_init(dateTimePickerWidgetClass *class)
...
+{
+	time_t t;
+	struct tm tmbuf;
+	GDateTime *dt;
+
+	dt = selected(d->d);
+	t = g_date_time_to_unix(dt);
+	g_date_time_unref(dt);
+
+	tmbuf = *localtime(&t);
```
andlabs: I wonder if combining these two lines instead of having the `tmpbuf` would be better. `localtime()` returns a pointer already, after all...

unix/datetimepicker.c
```diff
@@ -532,6 +541,68 @@ static void dateTimePickerWidget_class_init(dateTimePickerWidgetClass *class)
...
+	memcpy(time, &tmbuf, sizeof(struct tm));
+}
+
+void uiDateTimePickerSetTime(uiDateTimePicker *d, const struct tm *time)
+{
+	time_t t;
+	struct tm tmbuf;
+
+	memcpy(&tmbuf, time, sizeof(struct tm));
```
andlabs: Add a comment here explaining why we are copying `time` (because `mktime()` modifies its argument).

windows/datetimepicker.cpp
```diff
@@ -131,6 +134,72 @@ static void uiDateTimePickerMinimumSize(uiWindowsControl *c, int *width, int *he
...
+	uiDateTimePicker *d = uiDateTimePicker(c);
+
+	if (nmhdr->code != DTN_DATETIMECHANGE)
+		return FALSE;
+	(*(d->onChanged))(d, d->onChangedData);
+	*lResult = 0;
+	return TRUE;
+}
+
+static void fromSystemTime(LPSYSTEMTIME systime, struct tm *time)
```
andlabs: Style consistency nit: don't use `LP` names (use `SYSTEMTIME *` instead)

windows/datetimepicker.cpp
```diff
@@ -131,6 +134,72 @@ static void uiDateTimePickerMinimumSize(uiWindowsControl *c, int *width, int *he
...
+{
+	ZeroMemory(time, sizeof(struct tm));
```
andlabs: Style consistency nit: space after `sizeof`

windows/datetimepicker.cpp
```diff
@@ -131,6 +134,72 @@ static void uiDateTimePickerMinimumSize(uiWindowsControl *c, int *width, int *he
...
+	time->tm_mday = systime->wDay;
+	time->tm_mon = systime->wMonth - 1;
+	time->tm_year = systime->wYear - 1900;
+	time->tm_wday = systime->wDayOfWeek;
+	time->tm_isdst = -1;
+}
+
+static void toSystemTime(const struct tm *time, LPSYSTEMTIME systime)
+{
+	ZeroMemory(systime, sizeof(SYSTEMTIME));
```
andlabs: Style consistency nit: space after `sizeof`

windows/datetimepicker.cpp
```diff
@@ -131,6 +134,72 @@ static void uiDateTimePickerMinimumSize(uiWindowsControl *c, int *width, int *he
...
+static void defaultOnChanged(uiDateTimePicker *d, void *data)
+{
+	// do nothing
+}
+
+void uiDateTimePickerTime(uiDateTimePicker *d, struct tm *time)
+{
+	SYSTEMTIME systime;
+
+	if (SendMessageW(d->hwnd, DTM_GETSYSTEMTIME, 0, (LPARAM) &systime) != GDT_VALID)
```
andlabs: Style consistency nit: `(LPARAM) (&systime)` (this is probably unnecessary but I'm never sure how this works)

windows/datetimepicker.cpp
```diff
@@ -131,6 +134,72 @@ static void uiDateTimePickerMinimumSize(uiWindowsControl *c, int *width, int *he
...
+		logLastError(L"error getting date and time");
+	fromSystemTime(&systime, time);
+}
+
+void uiDateTimePickerSetTime(uiDateTimePicker *d, const struct tm *time)
+{
+	SYSTEMTIME systime;
+
+	toSystemTime(time, &systime);
+	if (!SendMessageW(d->hwnd, DTM_SETSYSTEMTIME, GDT_VALID, (LPARAM) &systime))
```
andlabs: Style consistency nits: same parentheses issue as before; compare with `== 0` instead of `!`

windows/datetimepicker.cpp
```diff
@@ -131,6 +134,72 @@ static void uiDateTimePickerMinimumSize(uiWindowsControl *c, int *width, int *he
...
+		logLastError(L"error setting date and time");
+	(*(d->onChanged))(d, d->onChangedData);
```
andlabs: `Set` functions should not call event handlers (yet).

ui.h
```diff
@@ -229,8 +229,12 @@ _UI_EXTERN void uiRadioButtonsSetSelected(uiRadioButtons *r, int n);
 _UI_EXTERN void uiRadioButtonsOnSelected(uiRadioButtons *r, void (*f)(uiRadioButtons *, void *), void *data);
 _UI_EXTERN uiRadioButtons *uiNewRadioButtons(void);
 
+struct tm;
 typedef struct uiDateTimePicker uiDateTimePicker;
 #define uiDateTimePicker(this) ((uiDateTimePicker *) (this))
+_UI_EXTERN void uiDateTimePickerTime(uiDateTimePicker *d, struct tm *time);
```
andlabs: Should probably add a TODO comment that the documentation should explain what happens with `tm_wday`, `tm_yday`, and `tm_isdst` (which I am leaving up to you if you have a better idea than [my suggestion](https://github.com/andlabs/libui/issues/295#issuecomment-366853845), since your code is inconsistent about this).

Also starting to lean toward including `<time.h>` instead of declaring `struct tm`, but I'll leave that up for later.

andlabs:
No, there are some non-nits in there...

cody271:
Okay, I will take another look at this sometime then..

ui.h
cody271: `tm_wday` and `tm_yday` will be undefined, and `tm_isdst` will always be -1 for now. I will document them as such. 

I believe `<time.h>` is fairly portable here, but nothing really wrong with that single forward declaration..

windows/datetimepicker.cpp
cody271: Agreed, need some sort of "properties" API long term to do this correctly. In this case though, this is simply how the control works. We call the event handler manually on Windows because its done for us on the other platforms. My example program makes straight forward use of this functionality.

darwin/datetimepicker.m
```diff
@@ -4,17 +4,120 @@
...
+	// do nothing
+}
+
+void uiDateTimePickerTime(uiDateTimePicker *d, struct tm *time)
+{
+	time_t t;
+	struct tm tmbuf;
+	NSDate *date;
+
+	date = [d->dp dateValue];
```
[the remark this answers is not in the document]
cody271: What's wrong with using `time_t` here?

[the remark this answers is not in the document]
andlabs: Oh, I was mostly concerned about wheher the type of `time_t` can be relied on on macOS, and to a lesser extent, the precision of NSDateInterval compared to that of time_t. Maybe I'm worrying too much and this code is fine as it is for now...

ui.h
andlabs: Fine by me 👍

darwin/datetimepicker.m
```diff
@@ -4,17 +4,120 @@
...
+
+void uiDateTimePickerSetTime(uiDateTimePicker *d, const struct tm *time)
+{
+	time_t t;
+	struct tm tmbuf;
+
+	memcpy(&tmbuf, time, sizeof(struct tm));
+	t = mktime(&tmbuf);
+
+	[d->dp setDateValue:[NSDate dateWithTimeIntervalSince1970:t]];
```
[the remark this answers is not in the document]
andlabs: (Also I'm just going to say that after several months of coding at Google with its code review tools (similar to public ones like Gerrit) that github's code review UI is really weird to me XD )

darwin/datetimepicker.m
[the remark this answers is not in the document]
andlabs: (Also I'm just going to say that after several months of coding at Google with its code review tools (similar to public ones like Gerrit) that github's code review UI is really weird to me XD )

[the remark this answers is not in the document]
andlabs: Also oops I meant to say accuracy not precision there. Still I'll look into it separately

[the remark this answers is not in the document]
andlabs: All right, after talking on #macdev it seems I am indeed paranoid (at least for the next 10,000 years), so yeah, this code will do for now.

Now how do I tell github that no action is needed...

darwin/datetimepicker.m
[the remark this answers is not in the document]
andlabs: No action needed here for now either.

andlabs:
Actually I just realized why you were copying the result of `mktime()` — to minimize a race condition because the time.h functions are not thread-safe, gah... I'll worry about that myself later though.

cody271:
Fixed with fd3b693.

andlabs:
Cool. No need for `struct uiprivMap` though; one of the new naming notes I need to add is to drop those from shared structs (I'll fix that myself).

Thanks again for all this! I'll merge everything tomorrow morning, as it's quite late. Good night!

PR #299, opened 2018-02-28
Update the Darwin screenshot to feature the new control gallery.

DanilaFe:
Since the new control gallery doesn't have any menu items on Darwin, I did not include the status bar in the screenshot.


parro-it:
BTW, the new control gallery is really nice!

andlabs:
What I'm not sure about is whether having only a subset of the controls visible in the screenshot is good enough...

As for this change, I was going to hold off until just before the alpha 4 release, because technically this new controlgallery is incomplete. (Thanks for the positive feedback about it, though!)

parro-it:
> What I'm not sure about is whether having only a subset of the controls visible in the screenshot is good enough...

On one side, I think the new window is more attractive, so I'll prefer it. I don't think it should necessarily contains all libui supported stuff. Just my 2c

DanilaFe:
I think it might be a good idea to have several screenshots for each platform on the README, though that might start taking up a lot of space.

PR #316, opened 2018-03-24
Resizeable Window

mischnic:
Add `uiWindowSetResizeable` (behaves like `uiWindowSetMargined`).

On Unix, setting to false makes the window go back to the initial window size (blame GTK).


sawyersteven:
Is there any plans to merge this? It seems like a rather basic feature that should be implemented.

PR #330, opened 2018-04-02
Added the LibUISharp project to the list of bindings.

tom-corwin:
This justs adds my C# bindings to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ Language | Bindings
 --- | ---
 C++ | [libui-cpp](https://github.com/billyquith/libui-cpp), [cpp-libui-qtlike](https://github.com/aoloe/cpp-libui-qtlike)
 C# / .NET Framework | [LibUI.Binding](https://github.com/NattyNarwhal/LibUI.Binding)
-C# / .NET Core | [DevZH.UI](https://github.com/noliar/DevZH.UI), [SharpUI](https://github.com/benpye/sharpui/)
+C# / .NET Core | [DevZH.UI](https://github.com/noliar/DevZH.UI), [SharpUI](https://github.com/benpye/sharpui/), [LibUISharp](https://github.com/tom-corwin/LibUISharp)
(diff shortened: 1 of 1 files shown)


andlabs:
Thanks!

I'm not sure why I am still separating the different .net frameworks...

tom-corwin:
@andlabs  .NET Framework and .NET Core (or libraries targeting .NET Standard) are the same, but different. Most notably, Microsoft's .NET Framework is [patented](https://github.com/Microsoft/referencesource/blob/master/PATENTS.txt)(https://github.com/Microsoft/reference-source/blob/master/LICENSE) and Windows only, where their [.NET Core](https://github.com/dotnet/core) is Microsoft's cross-platform, open-source alternative. (see [here](https://docs.microsoft.com/en-us/dotnet/standard/choosing-core-framework-server))

The .NET Framework also has a sister implementation, [mono](https://github.com/mono/mono) which runs on Windows, macOS, and Linux.

But, IMHO, I don't think they need to be separated, as the bindings organized by language, not SDK, plus you can cross-compile C# for both frameworks pretty easily.

andlabs:
What I want to eventually do is expand the table to have one row per binding with some extra information in a third column (right now the only thing I can think of is the licensing, since I noticed some of the language bindings are licensed differently from libui itself). I could move the .net framework compatibility level information there too. In particular, I'm still not sure if there's any software that can be used with one but not the other that don't do anything beyond the basic `System.*` namespaces that everyone learns about when they first learn a .net language.

I mostly made that remark based on some other peoples's responses to looking at that table, but I have to pull those responses back out first.

tom-corwin:
I'm not sure if it will help, but this, in the most basic sense, explains the correlation between the two frameworks, which shows that Core and Framework are separate APIs for different uses, built on top of a shared, standard API:
<sub>[Image Citation](https://www.slideshare.net/aminmesbahi/net-core-20-aspnet-core-20-entity-framework-core-20-overview)</sub>

PR #332, opened 2018-04-03
Add fixed positioning, and the ability to set and get a component size

### LINKED ISSUE #306: Fixed positioning
Hi,

For `proton-native`, I want to incorporate Yoga Layout, which I can do on the JS side without adding extra dependencies here, but I would need a way to fixed position place widgets using x and y coordinates. I've been playing around with GTK, and I got it working here: https://github.com/kusti8/libui, but I don't know how to do that for Windows or for Mac. This would be more of a substitute for Grid and make it a lot easier to use and more stable.

Thanks,
Gustav

kusti8:
This fixes #306 and allows me to get yoga working. So far, I've added this and the libui-node bindings as well as yoga to proton-native and it works very well.

I haven;t done any sort of work with GUI libraries really, so feel free to comment on my style. This shouldn't interfere with anything else, since all it does is add a new component. I also did this from the alpha3.5 tag because that is the latest that libui-node supported at the time. It is currently being worked on to work on the latest version.

Gustav


ui.h
```diff
@@ -97,6 +97,8 @@ _UI_EXTERN void uiControlHide(uiControl *);
 _UI_EXTERN int uiControlEnabled(uiControl *);
 _UI_EXTERN void uiControlEnable(uiControl *);
 _UI_EXTERN void uiControlDisable(uiControl *);
+_UI_EXTERN void uiSize(uiControl *control, int *width, int *height);
+_UI_EXTERN void uiSetSize(uiControl *control, int width, int height);
```
tom-corwin: For consistency, should these be names uiControlSize and uiControlSetSize, respectively?

darwin/fixed.m
```diff
@@ -0,0 +1,201 @@
+// 15 august 2015
+#import "uipriv_darwin.h"
+
+// TODO hiding all stretchy controls still hugs trailing edge
```
andlabs: Does this TODO have anything to do with this file?

.gitignore
```diff
@@ -0,0 +1,2 @@
+build
```
andlabs: Please remove this file.

darwin/fixed.m
```diff
@@ -0,0 +1,201 @@
...
+
+@interface fixedChild : NSObject
```
andlabs: Use the name `uiprivFixedChild`.

darwin/fixed.m
```diff
@@ -0,0 +1,201 @@
...
+@property uiControl *c;
+@property int x;
+@property int y;
+- (NSView *)view;
+@end
+
+@interface fixedView : NSView {
```
andlabs: Use the name `uiprivFixedView`.

darwin/fixed.m
```diff
@@ -0,0 +1,201 @@
...
+	uiFixed *b;
+	NSMutableArray *children;
+}
+- (id)initFixed:(uiFixed *)bb;
+- (bool)isFlipped;
```
andlabs: No need to redeclare `isFlipped`.

darwin/fixed.m
```diff
@@ -0,0 +1,201 @@
...
+}
+
+@end
+
+@implementation fixedView
+
+- (id)initFixed:(uiFixed *)bb
+{
+	self = [super initWithFrame:NSZeroRect];
+	self.autoresizingMask = NSViewWidthSizable | NSViewHeightSizable;
```
andlabs: Always use Objective-C method syntax, even for superclass property getters and setters.

Put this line in the if block.

libui uses Auto Layout; `uiprivFixedView` should be controllable with Auto Layout, but its contents don't necessarily have to be. I have to figure out how this would work.

darwin/fixed.m
```diff
@@ -0,0 +1,201 @@
...
+	if (self != nil) {
+		// the weird names vert and bb are to shut the compiler up about shadowing because implicit this/self is stupid
```
andlabs: No need to copy this comment over from the other files.

darwin/main.m
```diff
@@ -10,6 +10,41 @@
 static BOOL (^isRunning)(void);
 static BOOL stepsIsRunning;
 
+@interface sizeWrapper : NSView {
```
andlabs: Why is any of this necessary? You also are not properly initializing or uninitializing instances of these objects.

darwin/main.m
```diff
@@ -243,3 +278,16 @@ void uiQueueMain(void (*f)(void *data), void *data)
...
+	CGRect size = [[[sizeWrapper alloc] addControl:control] getSize];
+	*width = size.size.width;
+	*height = size.size.height;
+}
+
+void uiSetSize(uiControl *control, int width, int height)
+{
+	CGSize s = CGSizeMake(width, height);
+	[[[sizeWrapper alloc] addControl:control] setS:s];
+}
```
andlabs: Please use newlines at the end of files.

unix/fixed.c
```diff
@@ -0,0 +1,87 @@
...
+
+struct fixedChild {
+	uiControl *c;
+};
+
+struct uiFixed {
+	uiUnixControl c;
+	GtkWidget *widget;
+	GtkContainer *container;
+	GtkFixed *fixed;
```
andlabs: Don't use GtkFixed; it does not properly theme with CSS. GtkLayout would be better, but the ideal solution would be to have a custom GtkContainer subclass that does the barest minimum possible. Look at the version of package ui that immediately predated libui for details. (I can do this later too.)

windows/winpublic.cpp
```diff
@@ -35,6 +35,16 @@ void uiWindowsEnsureMoveWindowDuringResize(HWND hwnd, int x, int y, int width, i
 		logLastError(L"error moving window");
 }
 
+void uiWindowsResizeWindow(HWND hwnd, int width, int height) {
```
andlabs: I'm pretty sure these two are provided in winutil.cpp anyway.

Will review the rest of the code later.

darwin/fixed.m
```diff
@@ -0,0 +1,201 @@
...
+		self->b = bb;
+		self->children = [NSMutableArray new];
+	}
+
+	return self;
+}
+
+- (bool)isFlipped
```
andlabs: `BOOL`, not `bool`.

darwin/fixed.m
kusti8: No. I copied box.m and then modified that since I was unfamiliar with the layout and I forgot to take that out.

darwin/fixed.m
kusti8: I don't think the autoresizingMask is strictly necessary, I think I just added that when trying to debug why the components weren't being sized correctly.

unix/fixed.c
kusti8: I'll look at that.

kusti8:
I've made most of the changes. As for having uiFixed control the sizing, I agree that it's a good idea, and it seems like it will work with my code so I'll change that.

darwin/fixed.m
```diff
@@ -165,7 +162,7 @@ static void uiFixedSyncEnableState(uiDarwinControl *c, int enabled)
 uiDarwinControlDefaultHugsBottom(uiFixed, view)
 uiDarwinControlDefaultChildEdgeHuggingChanged(uiFixed, view)
 
-static void uiFixedChildVisibilityChanged(uiDarwinControl *c)
+static void uiuiprivFixedChildVisibilityChanged(uiDarwinControl *c)
```
mischnic: One "ui" too much

kusti8: Yeah I pushed so that I could check the changes in my VM. I'll fix that.

darwin/main.m
kusti8: Fixed.

windows/winpublic.cpp
kusti8: There was `uiWindowsEnsureMoveWindowDuringResize`, but that moves and resizes the window. I added these functions because when you either move or resize a window, Windows doesn't give accurate values for the other two, so if you want it to move, it'll end up resizing also. Which is why I made these with the `SWP_NOSIZE` and `SWP_NOMOVE` flags.

andlabs:
Also I need to formalize some of the suggestions I made here into the contribution guidelines. Sorry if I came off as aggressive with any of them; these are just policies specific to libui to keep everything consistent!

kusti8:
No it's all right. I'm all for keeping code quality and style consistent.
I've just never really worked with Objective C and C++ as much.

kusti8:
Alright sizing should be moved to uiFixed now.

unix/fixed.c
kusti8: Ok I finally got more time to look at this. I assume you're talking about this: https://github.com/andlabs/ui/blob/de9d72299fb89a8b6cdc8963cd6b6ae708a81e80/container_unix.c
So you're overriding GtkContainer, but where do you actually set children size and location?

andlabs: Line 62, which is a Go function. I forget what you call in there for each child control, but you can check the Go code.

kusti8:
@andlabs I know you're busy but what's the status on this?

brunolemos:
Hopefully this can get merged so the work on https://github.com/kusti8/proton-native/pull/76 can continue!

andlabs:
I was going to write a big long expalantion of why I didn't want to merge this PR in, but thinking about it now it's been too long and I forgot exactly what I was going to say at the time. Sorry.

Here's the general gist of what my objections were:

- the idea of the fixed layout in the first place is one I deliberately wanted to avoid having at all, because of the room for error and the amount of work it would take to make it error-free
- I didn't want to include Yoga in libui itself as a fix for uiGrid, partially because the license is not compatible, partially because it would be a massive dependency, partially because GtkGrid, and partially because Auto Layout
- I still don't fully understand Auto Layout — it seems like things like responding to when a basic control's intrinsic content size has changed should be something it does for us, and yet many of the bugs in uiBox and uiGrid can be boiled down to "it's not that simple"; at the same time, this PR doesn't seem to watch intrinsic content size changes at all

Of course, these original comments seem invalid now anyway, because I only see the Fixed code in the PR, not the code for any of the support APIs. What happened?

And then again, other UI toolkits have had fixed layouts alongside dynamic layouts for a while and haven't had major problems, so maybe I was just worrying too much. Perhaps a uiFixed is fine, though there's still a few things missing from libui that would allow one to be used "correctly".

kusti8:
That's alright. 

The original PR never included any yoga or anything like that for the exact purpose to keep libui light weight. I simply added an API for fixed position of objects. That API is used by me in JS which integrates with Yoga at that level, since yoga had a JS wrapper. Whever a change is seen by me, that is sent to yoga to recompute the coordinates and that moves everything.

andlabs:
The ideal thing is there should be one place where you can hook into to spot control size changes. Alpha 5 will include that, since I'll be redoing uiControl entirely. This should probably also automatically mirror on an RTL system, but I have no idea where to start with that.

SkyzohKey:
Hey @andlabs @kusti8 any progress on this ? :)

langovoi:
Any updates?

andlabs:
The work on Alpha 5 will allow this to be feasible without encouraging the kind of mistakes that cause other toolkits (including this one) to avoid including this. It'll have to wait one more version.

PR #157, opened 2016-06-18
Automating binary attachments to release.

parro-it:
Hi, I'm trying to automate the process of binaries attachment to GH releases.
We previously discussed this stuff on #124.

I will use these binaries on [libui-download](https://github.com/parro-it/libui-download) to avoid compile libui on each libui-node install.

This is a work in progress. Just publish the PR to get early feedback from you...

Some open issues and questions:
- You can see the actual attached files [on this test release](https://github.com/parro-it/libui/releases/tag/test13). Please suggest improvement or changes to file names as you like.
- I created many tags on my fork to trigger travis release builds, I don't know if these could be polluting this PR. please double check!
- The travis yaml file is becoming a mess. I'm thinking of creating a scripts folder to move some commands there... thoughts?
- Actually, there is no binaries for windows and for linux 32bit. I could try [to use docker](https://github.com/travis-ci/travis-ci/issues/5770) on travis to get a 32bit env, and to use appveyor for windows build
- You'll have to change the secure key with one for your repo after merge the PR. The actual one it's for my fork...

PR #340, opened 2018-04-19
Fix uiprivTimer::f,  it should return an int

parro-it:
On Windows, f is declared as a void returning function:

https://github.com/andlabs/libui/blob/591b9c87721b2e38fe6b2b1c455187375c764fa6/windows/uipriv_windows.hpp#L98-L102

But here it's result is negated as it's returning an int:

https://github.com/andlabs/libui/blob/591b9c87721b2e38fe6b2b1c455187375c764fa6/windows/utilwin.cpp#L40-L46

And here, it's set an int returning function:

https://github.com/andlabs/libui/blob/591b9c87721b2e38fe6b2b1c455187375c764fa6/windows/main.cpp#L132-L142

See also discussion here: https://github.com/andlabs/libui/pull/157

### DIFF
--- a/windows/uipriv_windows.hpp
+++ b/windows/uipriv_windows.hpp
@@ -97,7 +97,7 @@ extern void uninitUtilWindow(void);
 // main.cpp
 typedef struct uiprivTimer;
 struct uiprivTimer {
-	void (*f)(void *);
+	int (*f)(void *);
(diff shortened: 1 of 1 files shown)


andlabs:
Thanks, though I see another error that I'm surprised the compiler accepted...

PR #353, opened 2018-04-29
Add uiBoxChild

e0ff:
Checks to see if a uiControl is a child of a uiBox


szanni:
Is there a specific reason for introducing this new function? To my understanding `uiControlParent(child) == uiControl(box)` does the exact same thing without introducing a new API call.

PR #287, opened 2018-01-05
Msys fixes

bcampbell:
These changes get the utflib-and-attrstr branch compiling under MinGW-w64 (for me anyway, YMMV ;- )

Possible fix for #259

PR #354, opened 2018-05-01
Fixes to compile on windows with msys2

bcampbell:
msys2 seems to be unsure of it's IDWriteTypography definition and
disables it by default. This overrides the disabling (The
IDWriteTypography definition looks OK to me, and seems to work
fine).

This commit also adds in a canary to detect compilers that have a non-16bit
wchar_t on windows (eg a gcc focused on porting unix code rather
than compiling windows code).

### DIFF
--- a/windows/utf16.cpp
+++ b/windows/utf16.cpp
@@ -1,6 +1,13 @@
 // 21 april 2016
 #include "uipriv_windows.hpp"
 
+// sanity check - make sure wchar_t is 16 bits (the assumption on windows)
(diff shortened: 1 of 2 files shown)


bcampbell:
just cleaned up version of https://github.com/andlabs/libui/pull/287

andlabs:
I'll take this as it is for now, with some stylistic tweaks and TODOs added later. Thanks!

A lot of my comments on the upcoming PRs will be style consistency things; I'm not making these for your personal code, only for the code in libui, so everything looks consistent. Sorry! I will suggest the first one now: there should only be one newline between things, not two.

Issue #308, opened 2018-03-23
Master issue for static library nonsense: Windows resource file copying and Unix build oddities


andlabs:
libui needs to be available as both a static and as a dynamic library, as different users have different requirements. Static libraries have two major issues:

- they share the same symbol namespace with whatever they are linked to
- on Windows, static libraries can't have resources

Right now, libui tries to hack around all this by a) obfuscating non-public symbols using a trick deduced by @pcwalton b) manually copying the compiled `resources.o` file on Windows into the `build\out` folder so it can be linked into binaries.

These options have caused nothing but build issues and maintenance nightmares for everyone, including myself, so I'm going to be fixing this properly:

- prefix all non-public names that aren't `static` with `uipriv`
- embed compiled resources on Windows directly into the binary somehow

I need to figure out whether all structs and classes (including ones only used in one source file) need to have this too. All GObject and Objective-C classes also need this change, because they share an application-global namespace regardless.

I need to figure out how to do the Windows resource precompiling thing in a way suitable for each toolchain.

mischnic:
The macOS linker warning (`built for newer OSX version (10.11) than being linked (10.8)`) should already be fixed with this:
https://github.com/andlabs/libui/blob/51f72abba84e3941deba1c8f699046a48d1f7fb9/CMakeLists.txt#L19

andlabs:
Oh good, so it's already fixed then because package ui on OS X uses a cmake build with that patch applied. Either way, those issues can be closed.

DemiMarie:
I think we should localize symbols in static libraries where we can.

andlabs:
What do you mean?

DemiMarie:
@andlabs I think we should keep using `objcopy --localize-hidden` whenever it is available on the build machine.  CMake can easily check for this.

andlabs:
Why?

For the record, the current static library hacks (all of them combined) doesn't properly inherit object options like `-fPIC`, relies on things having specific pathnames that have proven wrong many times, and the resultant `.a` file doesn't get rebuilt when the source files change, which is annoying for testing.

andlabs:
Windows resource copying is done.

andlabs:
Oops, forgot to mark unix-namespace-cleanup as part of this issue.

Most of the immediate work is now finished. I still need to finalize naming policy and apply these fixes to Windows, but we should be better off than before in the meantime. I'll probably make this stop blocking Alpha 4's release now...

PR #358, opened 2018-05-04
CI: use AppVeyor for testing Windows build, upload artifacts from both Travis and AppVeyor

msink:


.appveyor.yml
```diff
@@ -0,0 +1,46 @@
...
+      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2013
+
+    - BUILD_TYPE: 386_shared
+      CMAKE_FLAGS: -G "Visual Studio 12 2013"
+      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2013
+
+    - BUILD_TYPE: amd64_mingw_static
+      MINGW_PATH: C:\msys64\mingw64\bin
+      CMAKE_FLAGS: -G "MinGW Makefiles" -DBUILD_SHARED_LIBS=OFF
+      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
```
parro-it: Why you use 2015 here?

parro-it: @andlabs should we use 2015 everywhere? Since it is the runtime included in [windows-build-tools](https://www.npmjs.com/package/windows-build-tools), it would require one less runtime library to inistall
for libui-node users. Actually, both 2013 and 2015 runtimes are required.

msink: Hm... I just always use 2015 in my projects. But yes, should work on 2013 and 2017 too.

msink: I can add any configuration here, but building every configuration takes around  one minute.
And in for free appveyor do not enables concurrent builds.

parro-it: I think we can just pick what @andlabs prefer.

parro-it:
Did you test the example exes?

msink:
No, this only first step - test buildability.
Next step - zipping artifacts.
I just do not know is it valid at all, let's wait for @andlabs

parro-it:
I notice the static version lib is huge, why?

.appveyor.yml
andlabs: compilerver.hpp has my minimum versions. Testing along the matrix of every version starting from it would be nice.

msink: added 2015 and 2017.
14 builds will run 20-25 minutes...

https://ci.appveyor.com/project/msink/libui/build/build%20%2321

msink: Well, actually it was 32 minutes.

CMakeLists.txt
```diff
@@ -81,7 +81,6 @@ if(MSVC)
 	set(_COMMON_CFLAGS
 		/W4 /wd4100
 		/bigobj /nologo
-		/RTC1 /RTCs /RTCu
```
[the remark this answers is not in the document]
andlabs: That is a separate PR I need to merge at some point.

andlabs:
I would put the shared libraries before the static libraries because they do not require any extra cmake arguments (and I prefer them anyway, because I could use resources; but that's gone now, so oh well).

So are we just abandoning @parro-it's work then?

msink:
So what exactly order you want? And maybe add `Debug` builds? It will be an hour...

About #157 - I actually take some pieces from it, so at least for now it needed.

parro-it:
> So are we just abandoning @parro-it's work then?

No worries, using the  Matrix functionality seems better.
@msink do you know if Travis has similar features?
Would you mind to also take care of that part?

msink:
Will try with Travis too, tomorrow.

But @parro-it - I see in your PR patches for building ia32 - do you really need ia32 or you confused it with i686? It not the same.

andlabs:
Do the Itanium versions of Windows Vista and Server 2008 have the Platform Update, and does the Itanium version of 2008 R2 have Direct2D? If so, then libui for Windows could run on it, but I wouldn't be able to test it... Would be cool to see running, though.

The order I'm imagining is something like

```
VS
  2013
    x86
      shared
      static
    x64
  2015
MinGW-w64
  (possibly other matrix axes here, IDK)
```

but that depends on how the AppVeyor matrix feature works...

parro-it:
> Will try with Travis too, tomorrow.

Awesome!

> But @parro-it - I see in your PR patches for building ia32 - do you really need ia32 or you confused it with i686? It not the same.

I confused it

msink:
@andlabs Sorry, but I still do not understand what build sequence you want :)

I think in our case it should be simple flat one-dimension matrix.
Because every build type requires different CMAKE_OPTIONS, and better keep it explicit in matrix, then build script is trivial and obvious.

Well, maybe add second dimension: `Debug/Release`, but I think `Debug` builds should not run every push/pull request, on tagged release - maybe, don't know.

So you want this build sequence:
```
x86_msvc2013_shared
x86_msvc2013_static
x86_msvc2015_shared
x86_msvc2015_static
x86_msvc2017_shared
x86_msvc2017_static
x64_msvc2013_static
x64_msvc2013_shared
x64_msvc2015_static
x64_msvc2015_shared
x64_msvc2017_static
x64_msvc2017_shared
x86_mingw_static
x64_mingw_static
```

Ot this:
```
x86_msvc2013_shared
x86_msvc2013_static
x64_msvc2013_static
x64_msvc2013_shared
x86_msvc2015_shared
x86_msvc2015_static
x64_msvc2015_static
x64_msvc2015_shared
x86_msvc2017_shared
x86_msvc2017_static
x64_msvc2017_static
x64_msvc2017_shared
x86_mingw_static
x64_mingw_static
```

And what resulting artifacts you and @parro-it want? Something like this loooong list:
```
x86_msvc2013_shared_debug.zip
x86_msvc2013_shared_release.zip
x86_msvc2013_shared_examples.zip
x86_msvc2013_static_debug.zip
x86_msvc2013_static_release.zip
x86_msvc2013_static_examples.zip
x86_msvc2015_shared_debug.zip
x86_msvc2015_shared_release.zip
x86_msvc2015_shared_examples.zip
x86_msvc2015_static_debug.zip
x86_msvc2015_static_release.zip
x86_msvc2015_static_examples.zip
x86_msvc2017_shared_debug.zip
x86_msvc2017_shared_release.zip
x86_msvc2017_shared_examples.zip
x86_msvc2017_static_debug.zip
x86_msvc2017_static_release.zip
x86_msvc2017_static_examples.zip
x86_mingw_static_debug.zip
x86_mingw_static_release.zip
x86_mingw_static_examples.zip   
x64_msvc2013_static_debug.zip
x64_msvc2013_static_release.zip
x64_msvc2013_static_examples.zip
x64_msvc2013_shared_debug.zip
x64_msvc2013_shared_release.zip
x64_msvc2013_shared_examples.zip
x64_msvc2015_static_debug.zip
x64_msvc2015_static_release.zip
x64_msvc2015_static_examples.zip
x64_msvc2015_shared_debug.zip
x64_msvc2015_shared_release.zip
x64_msvc2015_shared_examples.zip
x64_msvc2017_static_debug.zip
x64_msvc2017_static_release.zip
x64_msvc2017_static_examples.zip
x64_msvc2017_shared_debug.zip
x64_msvc2017_shared_release.zip
x64_msvc2017_shared_examples.zip
x64_mingw_static_debug.zip
x64_mingw_static_release.zip
x64_mingw_static_examples.zip   
```

Not that it's major issue - you always can reshuffle/rename anything youself - but shill.

andlabs:
Okay, so I'm not familiar with things. What is this AppVeyor matrix feature?

The order I was thinking of was

```
x86_msvc2013_shared
x86_msvc2013_static
x64_msvc2013_shared
x64_msvc2013_static
x86_msvc2015_shared
x86_msvc2015_static
x64_msvc2015_shared
x64_msvc2015_static
x86_msvc2017_shared
x86_msvc2017_static
x64_msvc2017_shared
x64_msvc2017_static
x86_mingw_static
x64_mingw_static
```

And I'm not sure what artifacts, hm... That would depend on how far a single artifact can be used. MinGW is a problem because of exception handling models, and I'm not sure what DLL artifacts can be used by which versions of MSVC. If any of you have answers to these questions, I'd love to hear them. Exception handling model is a bug at https://github.com/andlabs/ui/issues/281.

msink:
Updated Travis build, but now `linux_386_static` failed:
```
[ 71%] Linking C static library liblibui-temporary.a
[ 71%] Built target libui-temporary
[ 72%] Removing hidden symbols
ld: Relocatable linking with relocations from format elf32-i386 (/home/travis/build/andlabs/libui/build/liblibui-temporary.a(attribute.c.o)) to format elf64-x86-64 (libui-combined.o) is not supported
```

Don't know what is `liblibui-temporary` and `liblibui-temporary`, and why all that tampering needed.

msink:
@andlabs In this context artifact is part of distribution, basicslly what @parro-it wants in #157 split single file `alpha3.5.tgz` into multiple artifacts.

About mingw - well, it's really `msys2-mingw', maybe name is too general.

msink:
My plan is to somewhat rebase #157 on top ow this - it will become order of magnitude smaller and cleaner.
If @parro-it have time and willingness - he can do, if not - I can do myself, but later, when/if this will be merged.

andlabs:
liblibui-temporary and liblibui-combined is what I'm trying to get rid of in #308. It's now gone on macOS; GTK+ to come later. The problem is `-fvisibility-hidden` didn't apply to static libraries, but re-linking to remove hidden symbols fixed it; doing this was just too quirky and had other problems as well.

I intuited what artifacts are; I was asking what AppVeyor's matrix feature is. And I still wouldn't know what artifacts we actually need to provide, which was the question I thought you were asking earlier =P

And all right at merging the PRs; thanks for working on this to the both of you!

Another thing I wonder; I forgot if I asked this in the other PR — does Travis provide anything that AppVeyor does not? I'm only using Travis now because it was the first one someone recommended me (in https://github.com/andlabs/ui/pull/3).

parro-it:
😎 great work @msink!
Regarding libui-node needs, I just use these artifacts:
```
libui-shared-linux-ia32-alpha3.5-master-002.tar.gz
libui-shared-linux-x64-alpha3.5-master-002.tar.gz
libui-shared-osx-ia32-alpha3.5-master-002.tar.gz
libui-shared-osx-x64-alpha3.5-master-002.tar.gz
libui-shared-windows-ia32-alpha3.5-master-002.tar.gz
libui-shared-windows-x64-alpha3.5-master-002.tar.gz
```

The list is from last published release, that I tagged "alpha3.5-master-002".
It would be beautiful if you can keep the same naming convention, that would preserve me to change libui-download to use the new names.

> My plan is to somewhat rebase #157 on top ow this - it will become order of magnitude smaller and cleaner.

There's really something there that you don't already have here?

msink:
So for now I commented out 'linux_386_static' build until linking will be fixed,
What to do wiyh MSVC `/RTC1 /RTCs /RTCu` options - don't know.

Travis and AppVeyor are almost identical, but for linux+ios and windows respectively. So no competition, nice coexistance.

@parro-it I personally do not like `ia32/x64` naming, it confusing prone.
IMHO it should be `x86/x86_64` or `i386/x86_64`

msink:
@parro-it Ok, maybe it's me confused somewhat. Anyway, not a major issue :)

msink:
@andlabs Sorry but I cannot say much about matrix details on travis and appveyor.
Definitely they are a little different in details - different implementations of the same idea.
And both evolve in same direction, influencing each other.

andlabs:
I prefer Go's architecture naming: 386 and amd64. I already use it in the binary releases I make by hand, and would rather it stay with this.

msink:
AFAIK Intel 386 ISA on Linux have two different incompatible ABI's - i386 (old) and x32 (new, since kernel 3.4).
So I too prefer 386 over ia32 - less error prone.

andlabs:
Hopefully the static build will actually work once I get rid of the temporary and combined madness. If that's not the case, then well... :\

msink:
Rebased and reenabled `static-linux-386` build.
So now the only remaining blocking issue for this is MSVC mess with `/RTC1 /RTCs /RTCu` flags.

andlabs:
I will fix it once I figure out how to resolve the differences between the two PRs on that front.

.appveyor.yml
```diff
@@ -2,77 +2,51 @@ version: 'build #{build}'
...
-      )
+  - md build && cd build
+  - cmake -G "%CMAKE_GENERATOR%" %CMAKE_FLAGS% ..
+  - if %compiler%==mingw ( mingw32-make tester examples )
+    else ( msbuild libui.sln /t:Build /p:Configuration=Release /p:Platform=%platform% )
 
 after_build:
-    - dir /S build\out
+  - if %platform%==x64 ( set "platform=amd64" ) else ( set "platform=386" )
+  - echo set "target=%linking%-windows-%compiler%-%platform%"
```
andlabs: For consistency with the prior releases, I'd have the order be `windows-%platform%-shared` and `windows-%platform%-%compiler%-static`.

Also `%platform%` would probbaly make more sense if it was named `%arch%` instead.

Or does AppVeyor require these particular formats? Either way, I can change these myself if you want; I'll set up AppVeyor sometime today (if not tonight).

.travis.yml
```diff
@@ -1,23 +1,81 @@
...
-  - linux
-  - osx
+language: c
 
-# This makes us use Ubuntu 14 instead of 12
-dist: trusty
+matrix:
+  include:
+    - env:
+      - BUILD_TYPE: shared-linux-amd64
```
andlabs: And yes, for consistency with prior releases, this should be `linux-amd64-shared` (and likewise for the rest), but again, I'm not sure if github or something else is imposing a requirement on these names. Are they?

parro-it: > Are they?

No, you can call them as you prefer...

msink: I already refactored this :)

parro-it:
@andlabs, @msink it would be useful if artifacts could be compressed using .tar.gz format, as it is what I'm using now in automated download of binaries during libui-node installation.

.travis.yml
msink: Well, sorry - only appveyor was refactored for now.

msink:
Hm, not sure.
`.tar.gz` is not so common on Windows as `.zip`
But can do, let @andlabs decide.

msink:
> So the important question: how would I trigger artifact deployment here? I only need artifacts deployed when I tag a version.

Last two lines of `.appveyor.yml` are just for that.
You can do `git tag` and push locally, or just create release in GitHub web interface - anyway, if appveyor sees new tag, it will upload and attach all artefacts into release.
I personally prefer do it in web inteface.

For now I finished with appveyor, uploading from travis is missed.

parro-it:
> .tar.gz is not so common on Windows as .zip

For sure, but having two separate formats force me to add another dependency to libui-download to handle zip format, and to process the files in two different ways depending on the platform.

> Also I would personally have the libs and examples in the same archive too, but I can do this myself.

This also is not desirable for me, having the example in the same archive as the libraries 
increment the size of the download when user install libui-node, and I'll have to manually delete the example after the download.

A compromise could be to have some archive published just for the sake of bindings implementors, containing only library and headers, and all having the same format.
For my needs, I'll just use the artifacts for shared libraries, 64bit of windows, linux and macOS

andlabs:
Wat, so you are not shipping libui with libui-node? I can't think of any issue with that, but it does seem very unusual... do any of the other libui bindings do this?

parro-it:
Not at all, or I would have to include all the platform binaries in the package... at the time I start, counting the 32bit ver, that would have been 6 binaries to include, just to pick the one efffectively used in a machine.
This is a common pattern on npm.

msink:
@parro-it - maybe better way is just compile libui on travis+appveyor as I did in my simple Kotlin/Native bindings https://github.com/msink/klib-ui ?

parro-it:
I completely agree. So I'll just continue to publish from my fork for now...
We can discuss again when 1.0.0 will come 😁

msink:
Ok, Travis works too.
https://github.com/msink/libui/releases/tag/test-2

andlabs:
Excellent, thanks!

I'll worry about the artifact naming myself, I suppose. Shouldn't be too hard to fix.

I wonder if it would be safe to remove your API keys from the files before merging, lest either of us get yelled at by either AppVeyor or Travis's security team?...

msink:
Keys will be checked only when/if you try set `git tag` for release, for `git push` they will be just ignored.
So I think merging this is safe.

zhaozg:
Please update readme.md to show appveory status.

PR #361, opened 2018-05-05
 Add minimal uiTable implementation for windows

bcampbell:
(The first two commits are cherry-picked from the msys fixes - I needed them to build. The table changes are in a single commit)

This uses the win32 common controls listview to implement uiTable.
There are limitations:
 - It supports only a single TextPart per column.
 - ImagePart, CheckboxPart and ProgessBarPart are not implemented.
 - There is no support for cell coloring.
 - Cell editing is not implemented.
Some of these will be very hard to support using the standard
common control listview, and probably require an entire custom
listview.

### DIFF
--- a/test/page16.c
+++ b/test/page16.c
@@ -132,6 +132,8 @@ uiBox *makePage16(void)
 
 	uiTableSetRowBackgroundColorModelColumn(t, 3);
 
+	uiTableAppendTextColumn(t, "Numbers", 8);
+
 	tc = uiTableAppendColumn(t, "Buttons");
 	uiTableColumnAppendCheckboxPart(tc, 7, 0);
 	uiTableColumnAppendButtonPart(tc, 6, 1);
--- a/windows/image.cpp
+++ b/windows/image.cpp
@@ -1,0 +1,31 @@
+#include "uipriv_windows.hpp"
+// stubbed out windows image list implementation.
+// Required for uiTable control, but windows implemenation
+// doesn't currently have image support.
+
+struct uiImage {
+	double width;
+	double height;
+	// HIMAGELIST images;
+};
+
+uiImage *uiNewImage(double width, double height)
+{
+	uiImage *i;
+
+	i = uiprivNew(uiImage);
+	i->width = width;
+	i->height = height;
+	return i;
+}
+
+void uiFreeImage(uiImage *i)
+{
+	uiprivFree(i);
+}
+
+void uiImageAppend(uiImage *i, void *pixels, int pixelWidth, int pixelHeight, int pixelStride)
+{
+	// not implemented
+}
+
--- a/windows/table.cpp
+++ b/windows/table.cpp
@@ -1,14 +1,12 @@
 #include "uipriv_windows.hpp"
 
-static void uiTableDestroy(uiControl *c);
-static void uiTableMinimumSize(uiWindowsControl *c, int *width, int *height);
-static BOOL onWM_NOTIFY(uiControl *c, HWND hwnd, NMHDR *nm, LRESULT *lResult);
+//static void uiTableMinimumSize(uiWindowsControl *c, int *width, int *height);
 
 struct uiTable;
 
 struct uiTableModel {
 	uiTableModelHandler *mh;
-	std::vector<uiTable*> tables;
+	std::vector<uiTable *> tables;
 };
 
 struct uiTableColumn {
@@ -22,7 +20,7 @@ struct uiTable {
 	uiWindowsControl c;
 	uiTableModel *model;
 	HWND hwnd;
-	std::vector<uiTableColumn*> columns;
+	std::vector<uiTableColumn *> columns;
 };
 
 void *uiTableModelStrdup(const char *str)
@@ -52,27 +50,31 @@ void uiFreeTableModel(uiTableModel *m)
 
 void uiTableModelRowInserted(uiTableModel *m, int newIndex)
 {
-	LVITEM item;
+	LVITEMW item;
 
+	ZeroMemory(&item, sizeof (LVITEMW));
 	item.mask = 0;
 	item.iItem = newIndex;
 	item.iSubItem = 0; //?
 	for (auto t : m->tables) {
-		ListView_InsertItem( t->hwnd, &item );
+		if (SendMessageW(t->hwnd, LVM_INSERTITEM, 0, (LPARAM) (&item)) == (LRESULT) (-1))
+			logLastError(L"error calling LVM_INSERTITEM in uiTableModelRowInserted()");
 	}
 }
 
 void uiTableModelRowChanged(uiTableModel *m, int index)
 {
 	for (auto t : m->tables) {
-		ListView_Update( t->hwnd, index );
+		if (SendMessageW(t->hwnd, LVM_UPDATE, (WPARAM) index, 0) == (LRESULT) (-1))
+			logLastError(L"error calling LVM_UPDATE in uiTableModelRowChanged()");
 	}
 }
 
 void uiTableModelRowDeleted(uiTableModel *m, int oldIndex)
 {
 	for (auto t : m->tables) {
-		ListView_DeleteItem( t->hwnd, oldIndex );
+		if (SendMessageW(t->hwnd, LVM_DELETEITEM, (WPARAM) oldIndex, 0) == (LRESULT) (-1))
+			logLastError(L"error calling LVM_DELETEITEM in uiTableModelRowDeleted()");
 	}
 }
 
@@ -80,6 +82,7 @@ void uiTableColumnAppendTextPart(uiTableColumn *c, int modelColumn, int expand)
 {
 	uiTable *t = c->t;
 	int lvIndex = 0;
+	LVCOLUMNW lvc;
 
 	if (c->modelColumn >=0) {
 		return; // multiple parts not implemented
@@ -92,16 +95,17 @@ void uiTableColumnAppendTextPart(uiTableColumn *c, int modelColumn, int expand)
 			break;
 		}
 		if (candidate->modelColumn >= 0) {
-			++lvIndex;
+			lvIndex++;
 		}
 	}
-	
-	LV_COLUMN lvc;
+
+	ZeroMemory(&lvc, sizeof (LVCOLUMNW));
 	lvc.mask = LVCF_FMT | LVCF_WIDTH | LVCF_TEXT;	/* | LVCF_SUBITEM; */
 	lvc.fmt = LVCFMT_LEFT;
 	lvc.cx = 120;	// TODO
 	lvc.pszText = c->name;
-	ListView_InsertColumn(c->t->hwnd, lvIndex, &lvc);
+	if (SendMessageW(c->t->hwnd, LVM_INSERTCOLUMN, (WPARAM) lvIndex, (LPARAM) (&lvc)) == (LRESULT) (-1))
+		logLastError(L"error calling LVM_INSERTCOLUMN in uiTableColumnPartSetTextPart()");
 }
 
 void uiTableColumnAppendImagePart(uiTableColumn *c, int modelColumn, int expand)
@@ -138,33 +142,11 @@ void uiTableColumnPartSetTextColor(uiTableColumn *c, int part, int modelColumn)
 
 uiWindowsControlAllDefaultsExceptDestroy(uiTable)
 
-uiTable *uiNewTable(uiTableModel *model)
-{
-	uiTable *t;
-	int winStyle = WS_CHILD | LVS_AUTOARRANGE | LVS_REPORT | LVS_OWNERDATA | LVS_SINGLESEL;
-
-	uiWindowsNewControl(uiTable, t);
-	new(&t->columns) std::vector<uiTableColumn*>();		// (initialising in place)
-	t->model = model;
-	t->hwnd = uiWindowsEnsureCreateControlHWND(WS_EX_CLIENTEDGE,
-		WC_LISTVIEW,
-		L"",
-		winStyle,
-		hInstance,
-		NULL,
-		TRUE);
-	model->tables.push_back(t);
-	uiWindowsRegisterWM_NOTIFYHandler(t->hwnd, onWM_NOTIFY, uiControl(t));
-	ListView_SetExtendedListViewStyle(t->hwnd, LVS_EX_FULLROWSELECT | LVS_EX_LABELTIP);
-	// TODO: try LVS_EX_AUTOSIZECOLUMNS
-	int n = (*(model->mh->NumRows))(model->mh, model);
-	ListView_SetItemCountEx(t->hwnd, n, 0);
-	return t;
-}
-
 uiTableColumn *uiTableAppendColumn(uiTable *t, const char *name)
 {
-	uiTableColumn *c = uiprivNew(uiTableColumn);
+	uiTableColumn *c;
+   
+	c = uiprivNew(uiTableColumn);
 	c->name = toUTF16(name);
 	c->t = t;
 	c->modelColumn = -1;	// -1 = unassigned
@@ -182,25 +164,32 @@ static void uiTableDestroy(uiControl *c)
 {
 	uiTable *t = uiTable(c);
 	uiTableModel *model = t->model;
-	std::vector<uiTable*>::iterator it;
+	std::vector<uiTable *>::iterator it;
 
 	uiWindowsUnregisterWM_NOTIFYHandler(t->hwnd);
 	uiWindowsEnsureDestroyWindow(t->hwnd);
 	// detach table from model
-	for (it = model->tables.begin(); it != model->tables.end(); ++it) {
+	for (it = model->tables.begin(); it != model->tables.end(); it++) {
 		if (*it == t) {
 			model->tables.erase(it);
 			break;
 		}
 	}
 	// free the columns
-	for (auto col: t->columns) {
+	for (auto col : t->columns) {
 		uiprivFree(col->name);
 		uiprivFree(col);
 	}
-	t->columns.~vector<uiTableColumn*>();	// (created with placement new, so just call dtor directly)
+	t->columns.~vector<uiTableColumn *>();	// (created with placement new, so just call dtor directly)
 	uiFreeControl(uiControl(t));
 }
+
+// suggested listview sizing from http://msdn.microsoft.com/en-us/library/windows/desktop/dn742486.aspx#sizingandspacing:
+// "columns widths that avoid truncated data x an integral number of items"
+// Don't think that'll cut it when some cells have overlong data (eg
+// stupidly long URLs). So for now, just hardcode a minimum:
+#define tableMinWidth 107		/* in line with other controls */
+#define tableMinHeight (14*3)	/* header + 2 lines (roughly) */
 
 static void uiTableMinimumSize(uiWindowsControl *c, int *width, int *height)
 {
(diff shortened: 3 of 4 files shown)


andlabs:
Yes, I'll merge those branches shortly.

andlabs:
Done.

windows/image.cpp
```diff
@@ -0,0 +1,31 @@
+#include "uipriv_windows.hpp"
+// stubbed out windows image list implementation.
+// Required for uiTable control, but windows implemenation
+// doesn't currently have image support.
+
+struct uiImage {
+	double width;
+	double height;
+	// HIMAGELIST images;
```
andlabs: Not a review comment, just a bit of information: I don't think we'll be able to use HIMAGELIST here, since that deals with a set of images that have the same pixel sizes, while uiImage deals with a set of images that has different pixel sizes but the same physical size (for DPI independence). Don't worry about changing anything; I'll work on it later if you don't already have another PR with it.

bcampbell: Yep, I agree. Wasn't planning to go anywhere further with HIMAGELIST, Just stubbing out enough to get the tables compiling. and threw in the comment while I was in listview mode. I didn't spot the multi-size vs different-images distinction.

andlabs:
I merged the Windows fixes into the table branch now; not sure why it's not reflected here...

bcampbell:
I rebased my changes to try and keep the history clean. This is nudging against the outer limits of my git-fu, so fingers crossed :- )

andlabs:
Seems to have worked. Will review.

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+{
+	for (auto t : m->tables) {
+		ListView_Update( t->hwnd, index );
+	}
+}
+
+void uiTableModelRowDeleted(uiTableModel *m, int oldIndex)
+{
+	for (auto t : m->tables) {
+		ListView_DeleteItem( t->hwnd, oldIndex );
```
andlabs: Likewise here.

```c
if (SendMesageW(t->hwnd, LVM_DELETEITEM, (WPARAM) oldIndex, 0) == (LRESULT) (FALSE))
	logLastError(L"error calling LVM_DELETEITEM in uiTableModelRowDeleted()");
```

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+	item.iSubItem = 0; //?
+	for (auto t : m->tables) {
+		ListView_InsertItem( t->hwnd, &item );
+	}
+}
+
+void uiTableModelRowChanged(uiTableModel *m, int index)
+{
+	for (auto t : m->tables) {
+		ListView_Update( t->hwnd, index );
```
andlabs: Likewise here.

```c
if (SendMesageW(t->hwnd, LVM_UPDATE, (WPARAM) oldIndex, 0) == (LRESULT) (FALSE))
	logLastError(L"error calling LVM_UPDATE in uiTableModelRowDeleted()");
```

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+	for (auto candidate : t->columns) {
+		if (candidate == c) {
+			break;
+		}
+		if (candidate->modelColumn >= 0) {
+			++lvIndex;
+		}
+	}
+	
+	LV_COLUMN lvc;
```
andlabs: Consistency: declarations at the top

More important, though:

a) use `LVCOLUMNW` — `LV_COLUMN` is technically a deprecated name as of at least Vista if not earlier, and the `W` is consistency to explicitly use Unicode
b) you should call `ZeroMemory(&lvc, sizeof (LVCOLUMNW))` before setting any fields (partially consistency, partially correctness)

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+	lvc.mask = LVCF_FMT | LVCF_WIDTH | LVCF_TEXT;	/* | LVCF_SUBITEM; */
+	lvc.fmt = LVCFMT_LEFT;
+	lvc.cx = 120;	// TODO
+	lvc.pszText = c->name;
+	ListView_InsertColumn(c->t->hwnd, lvIndex, &lvc);
```
andlabs:
```c
if (SendMessageW(c->t->hwnd, LVM_INSERTCOLUMN, lvIndex, (LPARAM) (&lvc)) == (LRESULT) (-1))
	logLastError(L"error calling LVM_INSERTCOLUMN in uiTableColumnAppendTextPart()");
```

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+void uiTableColumnPartSetTextColor(uiTableColumn *c, int part, int modelColumn)
+{
+	// not implemented
+}
+
+// uiTable implementation
+
+uiWindowsControlAllDefaultsExceptDestroy(uiTable)
+
+uiTable *uiNewTable(uiTableModel *model)
```
andlabs: This should come after all the other uiTable methods, for consistency. This will also avoid the need to forward-declare the Destroy function as you did above.

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+{
+	uiTable *t;
+	int winStyle = WS_CHILD | LVS_AUTOARRANGE | LVS_REPORT | LVS_OWNERDATA | LVS_SINGLESEL;
```
andlabs: What consequence on the model does `LVS_AUTOARRANGE` have? I'm leaning toward removing it for now.

You will also need to throw in `WS_TABSTOP`, `WS_HSCROLL`, and `WS_VSCROLL`. Not sure if `LVS_SHOWSELALWAYS` should also be used; see the other controls for consistency?

You will not need `WS_CHILD` with `uiWindowsEnsureCreateControlHWND()`.

(Also ideally this should be embedded in the call to `uiWindowsEnsureCreateControlHWND()`; the variable can be used for flags that are optional, but in that case the type should be `DWORD`.

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+	uiWindowsNewControl(uiTable, t);
+	new(&t->columns) std::vector<uiTableColumn*>();		// (initialising in place)
+	t->model = model;
+	t->hwnd = uiWindowsEnsureCreateControlHWND(WS_EX_CLIENTEDGE,
```
andlabs: This should be consistent with the formatting of the rest of the files:

```c
 	t->hwnd = uiWindowsEnsureCreateControlHWND(WS_EX_CLIENTEDGE,
		WC_LISTVIEW, L"",
		LVS_REPORT | LVS_OWNERDATA | LVS_SINGLESEL | WS_TABSTOP | WS_HSCROLL | WS_VSCROLL,
		hInstance, NULL,
		TRUE);
```

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+		WC_LISTVIEW,
+		L"",
+		winStyle,
+		hInstance,
+		NULL,
+		TRUE);
+	model->tables.push_back(t);
+	uiWindowsRegisterWM_NOTIFYHandler(t->hwnd, onWM_NOTIFY, uiControl(t));
+	ListView_SetExtendedListViewStyle(t->hwnd, LVS_EX_FULLROWSELECT | LVS_EX_LABELTIP);
```
andlabs:
```c
SendMessageW(t->hwnd, LVM_SETEXTENDEDLISTVIEWSTYLE,
	(WPARAM) (LVS_EX_FULLROWSELECT | LVS_EX_LABELTIP),
	(LPARAM) (LVS_EX_FULLROWSELECT | LVS_EX_LABELTIP));
```

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+	// TODO: try LVS_EX_AUTOSIZECOLUMNS
+	int n = (*(model->mh->NumRows))(model->mh, model);
```
andlabs: Variables declared at the top

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+	ListView_SetItemCountEx(t->hwnd, n, 0);
```
andlabs:
```c
if (SendMessageW(t->hwnd, LVM_SETITEMCOUNT, (WPARAM) n, 0) == 0)
	logLastError(L"error calling LVM_SETITEMCOUNT in uiNewTable()");
```

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+	return t;
+}
+
+uiTableColumn *uiTableAppendColumn(uiTable *t, const char *name)
+{
+	uiTableColumn *c = uiprivNew(uiTableColumn);
```
andlabs: Consistency: separate the declaration from the assignment here.

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+void uiTableSetRowBackgroundColorModelColumn(uiTable *t, int modelColumn)
+{
+	// not implemented
+}
+
+static void uiTableDestroy(uiControl *c)
+{
+	uiTable *t = uiTable(c);
+	uiTableModel *model = t->model;
+	std::vector<uiTable*>::iterator it;
```
andlabs: Style consistency: space before `*` in pointer type

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+	uiWindowsUnregisterWM_NOTIFYHandler(t->hwnd);
+	uiWindowsEnsureDestroyWindow(t->hwnd);
+	// detach table from model
+	for (it = model->tables.begin(); it != model->tables.end(); ++it) {
```
andlabs: Style consistency: `it++`

Style consistency: no braces around single statement for body

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+		if (*it == t) {
+			model->tables.erase(it);
+			break;
+		}
+	}
+	// free the columns
+	for (auto col: t->columns) {
```
andlabs: Style consistency: space before `:` as well as after

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+	uiTable *t = uiTable(c);
+	uiWindowsSizing sizing;
+	int x, y;
+
+	// suggested listview sizing from http://msdn.microsoft.com/en-us/library/windows/desktop/dn742486.aspx#sizingandspacing:
+	// "columns widths that avoid truncated data x an integral number of items"
+	// Don't think that'll cut it when some cells have overlong data (eg
+	// stupidly long URLs). So for now, just hardcode a minimum:
+
+	x = 107;	// in line with other controls
```
andlabs: I would name these constants somehow, for consistency

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+			// make sure clipped strings are nul-terminated
+			if (n>=item->cchTextMax) {
+				item->pszText[item->cchTextMax-1] = L'\0';
+			}
+		} else {
+			item->pszText[0] = L'\0';
+		}
+		break;
+	}
+	default:
```
andlabs: No need for `default: break;`

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+static BOOL onWM_NOTIFY(uiControl *c, HWND hwnd, NMHDR *nm, LRESULT *lResult)
+{
+	uiTable *t = uiTable(c);
+	uiTableModelHandler *mh = t->model->mh;
+	BOOL ret = FALSE;
+
+	switch (nm->code) {
+	case LVN_GETDISPINFO:
+	{
+		NMLVDISPINFO* di = (NMLVDISPINFO*)nm;
```
andlabs: Style consistency:

```c
		NMLVDISPINFO *di = (NMLVDISPINFO *) nm;	
		LVITEM *item = &(di->item); 
```

Note that the rest of this function isn't consistent either, but we can worry about that later.

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+			break;
+		}
+
+		uiTableColumn *tc = (uiTableColumn*)t->columns[col];
+
+		int mcol = tc->modelColumn;
+		uiTableModelColumnType typ = (*mh->ColumnType)(mh,t->model,mcol);
+		if (typ == uiTableModelColumnString) {
+			void* data = (*(mh->CellValue))(mh, t->model, row, mcol);
+			int n = MultiByteToWideChar(CP_UTF8, 0, (const char*)data, -1, item->pszText, item->cchTextMax);
```
andlabs: This might be a problem. It should be using `toUTF16()`, but I forget what the behavior is if you overwrite the `pszText` pointer with your own... or when it's safe to free. Just mark this as a TODO; I'll probably resolve it a different way.

windows/table.cpp
```diff
@@ -0,0 +1,275 @@
...
+			// make sure clipped strings are nul-terminated
+			if (n>=item->cchTextMax) {
+				item->pszText[item->cchTextMax-1] = L'\0';
+			}
+		} else if (typ == uiTableModelColumnInt) {
```
andlabs: Just a note: this won't be used for strings, but will be for progressbars. I will need to explicitly document that... Mark that as a TODO for now, anyway; this code can stay for now.

bcampbell:
Just a quick note to assure you I'm still here ;-) It's been a crazy-busy week or so, so I've not been able to spend any time on this. The feedback is great and all makes sense to me, and I'm aiming to apply it all over the next few days.

andlabs:
Okay; good luck with everything else going on!

test/page16.c
```diff
@@ -132,6 +132,8 @@ uiBox *makePage16(void)
 
 	uiTableSetRowBackgroundColorModelColumn(t, 3);
 
+	uiTableAppendTextColumn(t, "Numbers", 8);
```
andlabs: As I mentioned in another comment, the integer column type is for progress-bar parts. not for text strings. I don't remember what happens with other platforms though, so keep this around and I'll remove it if it causes issues.

However, this does raise the question of whether to provide helpers for displaying numbers as text. That might be outside the scope of libui, but it's still an interesting question...

windows/table.cpp
```diff
@@ -0,0 +1,295 @@
+#include "uipriv_windows.hpp"
+
+//static void uiTableMinimumSize(uiWindowsControl *c, int *width, int *height);
+
+struct uiTable;
```
andlabs: No need for this; `ui_table.h` already does this.

windows/table.cpp
```diff
@@ -0,0 +1,295 @@
+#include "uipriv_windows.hpp"
+
+//static void uiTableMinimumSize(uiWindowsControl *c, int *width, int *height);
```
andlabs: No need for this; `ui_windows.h` already does this.

windows/table.cpp
```diff
@@ -0,0 +1,295 @@
...
+}
+
+void uiTableModelRowInserted(uiTableModel *m, int newIndex)
+{
+	LVITEMW item;
+
+	ZeroMemory(&item, sizeof (LVITEMW));
+	item.mask = 0;
+	item.iItem = newIndex;
+	item.iSubItem = 0; //?
```
andlabs: 0 appears to be correct here. yes, since a subitem index of 0 means the item itself, not any of its subitems. (The leftmost column is not considered as "subitem" because of how listview is basically an extended listbox.)

windows/table.cpp
```diff
@@ -0,0 +1,295 @@
...
+	for (auto t : m->tables) {
```
andlabs: Style consistency nit: no need for braces here since the body is one statement.

windows/table.cpp
```diff
@@ -0,0 +1,295 @@
...
+		if (SendMessageW(t->hwnd, LVM_INSERTITEM, 0, (LPARAM) (&item)) == (LRESULT) (-1))
+			logLastError(L"error calling LVM_INSERTITEM in uiTableModelRowInserted()");
+	}
+}
+
+void uiTableModelRowChanged(uiTableModel *m, int index)
+{
+	for (auto t : m->tables) {
```
andlabs: Style consistency nit: no need for braces here since the body is one statement.

windows/table.cpp
```diff
@@ -0,0 +1,295 @@
...
+		if (SendMessageW(t->hwnd, LVM_UPDATE, (WPARAM) index, 0) == (LRESULT) (-1))
+			logLastError(L"error calling LVM_UPDATE in uiTableModelRowChanged()");
+	}
+}
+
+void uiTableModelRowDeleted(uiTableModel *m, int oldIndex)
+{
+	for (auto t : m->tables) {
```
andlabs: Style consistency nit: no need for braces here since the body is one statement.

windows/table.cpp
```diff
@@ -0,0 +1,295 @@
...
+	}
+}
+
+void uiTableColumnAppendTextPart(uiTableColumn *c, int modelColumn, int expand)
+{
+	uiTable *t = c->t;
+	int lvIndex = 0;
+	LVCOLUMNW lvc;
+
+	if (c->modelColumn >=0) {
```
andlabs: Style consistency nits: space after `>=`, no need for braces here since the body is one statement. I'll stop this second one for the rest of the review, but it still applies.

windows/table.cpp
```diff
@@ -0,0 +1,295 @@
...
+	int x, y;
+
+	x = tableMinWidth;
+	y = tableMinHeight;
+	uiWindowsGetSizing(t->hwnd, &sizing);
+	uiWindowsSizingDlgUnitsToPixels(&sizing, &x, &y);
+	*width = x;
+	*height = y;
+}
+
```
andlabs: Style consistency nit applied to everything: use only a single blank line between functions. not two.

windows/table.cpp
```diff
@@ -0,0 +1,295 @@
...
+	t->columns.~vector<uiTableColumn *>();	// (created with placement new, so just call dtor directly)
+	uiFreeControl(uiControl(t));
+}
+
+// suggested listview sizing from http://msdn.microsoft.com/en-us/library/windows/desktop/dn742486.aspx#sizingandspacing:
+// "columns widths that avoid truncated data x an integral number of items"
+// Don't think that'll cut it when some cells have overlong data (eg
+// stupidly long URLs). So for now, just hardcode a minimum:
+#define tableMinWidth 107		/* in line with other controls */
+#define tableMinHeight (14*3)	/* header + 2 lines (roughly) */
```
andlabs: There might be a way to use `LVM_GETHEADER` to get the necessary size of the header control, perhaps combined with `HDM_LAYOUT`, but I'm not sure how this cooperates with listview controls. Mark this as a TODO for now.

windows/table.cpp
```diff
@@ -0,0 +1,295 @@
...
+		LVS_REPORT | LVS_OWNERDATA | LVS_SINGLESEL | WS_TABSTOP | WS_HSCROLL | WS_VSCROLL,
+		hInstance, NULL,
+		TRUE);
+	model->tables.push_back(t);
+	uiWindowsRegisterWM_NOTIFYHandler(t->hwnd, onWM_NOTIFY, uiControl(t));
+
+	SendMessageW(t->hwnd, LVM_SETEXTENDEDLISTVIEWSTYLE,
+		(WPARAM) (LVS_EX_FULLROWSELECT | LVS_EX_LABELTIP),
+		(LPARAM) (LVS_EX_FULLROWSELECT | LVS_EX_LABELTIP));
+	// TODO: try LVS_EX_AUTOSIZECOLUMNS
```
andlabs: Move this above the `SendMessage()` call.

test/page16.c
bcampbell: It seems to me that the type of the cell used in the table display is independent from datatype of the column in the model. While most mappings between the data and display types are clearly pretty bonkers (eg colour => progress bar doesn't sound too useful), it doesn't seem unreasonable for the table to have a good stab at the sane ones (one?). Particularly the numeric => text mapping.
It's a pretty common case to want to display numeric data in a table, so forcing the client app to present it's numeric data as text when the model already has perfectly good support for numeric data seems like a recipe for confusion...

bcampbell:
OK, I _think_ that addresses them all. Let me know if I missed any!

windows/table.cpp
```diff
@@ -130,7 +119,7 @@ void uiTableColumnAppendProgressBarPart(uiTableColumn *c, int modelColumn, int e
 
 void uiTableColumnPartSetEditable(uiTableColumn *c, int part, int editable)
 {
-	// TODO
```
andlabs: Keep this as a TODO for now, so that it shows up in my occasional `grep TODO`s and thus isn't forgotten.

andlabs:
All right; thanks! I'll merge this into table now at last, though there are a few other minor changes and one last runthrough before I merge this into master. Thanks a lot again!

PR #373, opened 2018-05-27
Some small, unimportant, stupid addition to README.md

p1v0t:
Added hyperlinks for CONTRIBUTIN.MD and ANNOUNCE.MD to README.md file


andlabs:
Thanks! Two suggestions, though:

1) use inline links (`[like](http://this/)`) to be consistent with the rest of the document
2) use relative paths (`[like this](CONTRIBUTING.md)`) since github will point you to the correct place automatically

andlabs:
You will also need to sync to HEAD.

PR #374, opened 2018-05-30
[DO NOT MERGE] fixes from melonDS

msink:
Patches currently applied to libui in melonDS
Maybe something from this should be applied to libui master?


andlabs:
Yes; please split this.

msink:
This is not my patches, I just cleaned and sorted them, no functionsl changes from me.
Anyway this cannot be merged "as-is" - in particular, support for OSX totally missed, melonDS does not work on Mac.

So what should I split?
Make 15 pull requests? Or some are definitely wrong and dont worth cleaning?

andlabs:
Are the 15 commits entirely different in scope? If so, then yeah; sorry ^^; If not, we can combine those that aren't into larger PRs.

And don't worry about OS X support; I can take care of that.

msink:
Ok, lets go step by step: #375 melonDS dialogs

> Also out of curiosity do you work on melonDS as well?
No, no relations.
So I do not want to make any functional changes, only formatting.

msink:
Well, splitted to 7 separate PRs.
Something remained here - I doubt it's applicable directly.

PR #375, opened 2018-05-31
melonDS: dialogs

msink:
fixes from melonDS related to dialog windows

### DIFF
--- a/darwin/stddialogs.m
+++ b/darwin/stddialogs.m
@@ -32,7 +32,7 @@ static char *runSavePanel(NSWindow *parent, NSSavePanel *s)
 	return filename;
 }
 
-char *uiOpenFile(uiWindow *parent)
+char *uiOpenFile(uiWindow *parent, char* filter)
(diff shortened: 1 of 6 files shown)


mischnic:
*My old comment*:

> add support for non-resizable windows

#316 has macOS support (but as a setter and not a constructor argument)

mischnic:
Is that file dialog `filter` a glob? 
Example from [melonDS](https://github.com/StapleButter/melonDS/blob/aab0030137d70a7ef06c2d2e859f195dd06e0079/src/libui_sdl/main.cpp#L808): `"DS ROM (*.nds)|*.nds;*.srl|Any file|*.*"`

Also, while we're at it: they have also initpath:
```c
_UI_EXTERN char *uiOpenFile(uiWindow *parent, char* filter, char* initpath);
```

msink:
So should I drop first patch, assuming #316 is better and will be merged instead?

PR #377, opened 2018-05-31
melonDS: fixes related to uiControl

msink:


mischnic:
Maybe `uiControlFocus` instead of `uiControlSetFocus`? "Set" gives the feeling the function is:
```c
void uiControlSetFocus(uiControl *c, int isFocused);
```

PR #378, opened 2018-05-31
melonDS: fixes related to uiWindow

msink:


andlabs:
I want to hold off on dropping files until a proper drag/drop system is implemented.

Not too happy with GetFocus/LoseFocus, as those names tend to refer to keyboard focus, not becoming the active window. Will need to review that part of the code again first.

PR #380, opened 2018-05-31
Do no draw uiSlider value on unix systems to match darwin & windows.

szanni:
As can be seen in the example screenshots provided in the README, the slider on Unix platforms looks somewhat different. It draws the current value above the slider. This commit fixes that - makes all platforms look the same.

### DIFF
--- a/unix/slider.c
+++ b/unix/slider.c
@@ -61,6 +61,9 @@ uiSlider *uiNewSlider(int min, int max)
 	s->range = GTK_RANGE(s->widget);
 	s->scale = GTK_SCALE(s->widget);
 
+	// do not display current value
+	gtk_scale_set_draw_value(s->scale, 0);
+
 	// ensure integers, just to be safe
 	gtk_scale_set_digits(s->scale, 0);
 


mischnic:
> This commit fixes that - makes all platforms look the same.

Maybe this could be added to windows and macos and exposed in the constructor?

andlabs:
I can take this if on GTK+ a tooltip is used instead.

mischnic:
> I forget if it's how it works on Mac.

On macOS there is no indication whatsoever.

mischnic:
This would show a tooltip with the value after hovering over the knob.

```diff
diff --git a/darwin/slider.m b/darwin/slider.m
index 6f5c5a76..8705a454 100644
--- a/darwin/slider.m
+++ b/darwin/slider.m
@@ -58,6 +58,7 @@ struct uiSlider {

        s = (uiSlider *) uiprivMapGet(self->sliders, sender);
        (*(s->onChanged))(s, s->onChangedData);
+       [s->slider setToolTip:[NSString stringWithFormat:@"%ld",[s->slider integerValue]]];
 }

 - (void)registerSlider:(uiSlider *)s
```

But more generally: What if the slider value isn't really what the user uses? Then the slider tooltip would be wrong.

szanni:
I just opened a new pr with a slightly different approach with default tooltips on all platforms and enable/disable functions to leave the choice up to the user in #424

PR #381, opened 2018-05-31
Add uiSliderSetRange function to API

szanni:
Add a function to the slider API to modify the slider range after slider creation.
This is useful for e.g. media players that reset the range with every track -> without having to resort to deleting the existing slider and creating a new one with all signals etc.

**THE WINDOWS CODE HAS NOT BEEN TESTED**
I sadly don’t have a windows machine at hand right now, but looking at the MSDN docs this should work.

I did not adhere completely to the style doc either - with regards to the temp variable, which is declared within the scope of the if block. Less lines, more screen estate. If this is ok I’d be willing to adjust the style guide to add this exception. If not here is the same content adhering strictly to the style guide:
https://github.com/szanni/libui/tree/uiSliderSetRange-formatting


mischnic:
You should set the slider value to be within the new range in `uiSliderSetRange`.

mischnic:
Check the appveyor CI, there is a build error on Visual Studio 2013:

```
"C:\projects\libui\build\libui.sln" (Build target) (1) ->
"C:\projects\libui\build\test\tester.vcxproj.metaproj" (default target) (15) ->
"C:\projects\libui\build\test\tester.vcxproj" (default target) (16) ->
(Link target) -> 
  page4.obj : error LNK2019: unresolved external symbol _snprintf referenced in function _onRangeSliderChanged [C:\projects\libui\build\test\tester.vcxproj]
  C:\projects\libui\build\out\Release\test.exe : fatal error LNK1120: 1 unresolved externals [C:\projects\libui\build\test\tester.vcxproj]
```

szanni:
Classic Windows not supporting the C standard. Apparently VS only started supporting the snprintf function as of 2015: https://social.msdn.microsoft.com/forums/vstudio/en-US/2b339bdf-7ab1-4a08-bf7e-e9293801455b/c99-standard-snprintf-function-

What is the best course of action here? The CONTRIBUTING.md states C++11 compatibility, snprintf is C99 so **should** be included. I know Microsoft thinks otherwise.

Should I change it to sprintf, make the buffer 12 char and assert() an Int has a max of 32 bits?

Or forget about the testcases compiling prior to VS 2015? Not sure if that would be the right approach.

andlabs:
Yeah, I ran into this omission once before in the tester. And yeah, the final version of C++11 does state `snprintf()` should be part of `<cstdio>`... Just copy what I did in that other place.

szanni:
Updated the branches. Should compile on VS2013 now too.

szanni:
Any chance of getting this merged? Anything blocking this pr?

PR #382, opened 2018-06-01
melonDS: fixes related to uiArea

msink:


andlabs:
Not a fan of that SetBackgroundColor(), especially since a) what it does can be done by a Draw handler, and b) it doesn't use the same floating-point RGBA system as everything else.

As for that comment about area handlers, I'd rather open an issue about replacing that + table model handlers with something better entirely. I'm not sure what's "better" though. It'd have to be a set-once thing.

Issue #331, opened 2018-04-02
uiDrawMatrixScale center


mischnic:
Are the `xCenter` and `yCenter` coordinates relative to the element or the area?
https://github.com/andlabs/libui/blob/9cf56e1674168689f677a803b3b59cb21a27e0cf/ui.h#L472

(I'm guessing it's relative to the area).
I'm trying to scale the rectangle with it's center as the scale center.
Here I drew the rectangle at (50,50) and set the scale center to (50,50). Why does the rectangle  move?

![uiscale](https://user-images.githubusercontent.com/4586894/38209048-54247c32-36b3-11e8-91b4-42266afa53f9.gif)

mischnic:
Complete example:
```c
#include <stdio.h>
#include <string.h>
#include "ui.h"

int x = 0;

static void handlerDraw(uiAreaHandler *a, uiArea *area, uiAreaDrawParams *p)
{

	uiDrawBrush brush;
	brush.Type = uiDrawBrushTypeSolid;
	brush.R = 255;
	brush.G = 0;
	brush.B = 0;
	brush.A = 1.0;

	uiDrawMatrix m;
	uiDrawMatrixSetIdentity(&m);
	uiDrawMatrixScale(&m, 50, 50, 0.5+x/300.0, 1);

	uiDrawPath *path = uiDrawNewPath(uiDrawFillModeWinding);
	uiDrawPathAddRectangle(path, 50, 50, 100, 100);
	uiDrawPathEnd(path);
	
	uiDrawTransform(p->Context, &m);

	uiDrawFill(p->Context, path, &brush);
	uiDrawFreePath(path);
}

static void handlerMouseEvent(uiAreaHandler *a, uiArea *area, uiAreaMouseEvent *e) {
	x = e->X;
	uiAreaQueueRedrawAll(area);
}
static void handlerMouseCrossed(uiAreaHandler *ah, uiArea *a, int left) {}
static void handlerDragBroken(uiAreaHandler *ah, uiArea *a) {}
static int handlerKeyEvent(uiAreaHandler *ah, uiArea *a, uiAreaKeyEvent *e) {return 0;}


int main(void)
{
	uiInitOptions o;
	const char *err;

	memset(&o, 0, sizeof (uiInitOptions));
	err = uiInit(&o);
	if (err != NULL) {
		fprintf(stderr, "error initializing ui: %s\n", err);
		uiFreeInitError(err);
		return 1;
	}

[...]

PR #383, opened 2018-06-01
melonDS: fix for drawmatrix rotation

msink:
…tion.


mischnic:
*My old comment*

> fix drawmatrix transform order to match Windows. fixes rotation.

Is that #331? Then macOS is wrong as well.

msink:
I'm not an expert in vector graphix, but with this path applied [this](https://github.com/msink/kotlin-libui/blob/master/samples/logo/main.kt) sample works on Linux.
If not - any non-zero rotation leads to blank Area.

PR #385, opened 2018-06-01
Fix uiAreaBeginUserWindowResize on macOS

### LINKED ISSUE #357: macOS uiAreaBeginUserWindowResize
The test's page 15 "Borderless Resize" doesn't for me with macOS Sierra.

Trying to resize the window on any side closes the window and prints:
```
2018-05-04 12:30:25.026 test[11922:4447363] This NSLayoutConstraint is being configured with a
 constant that exceeds internal limits.  A smaller value will be substituted, but this problem should be
 fixed. Break on BOOL _NSLayoutConstraintNumberExceedsLimit() to debug.  This will be logged only
 once.  This may break in the future.
```

mischnic:
1. `CGFLOAT_MAX` isn't valid value for a constraint (is 100000 is ok?)
2. The position (= origin) of the window changes when resizing on the left or bottom edge.

Fixes #357

### DIFF
--- a/darwin/winmoveresize.m
+++ b/darwin/winmoveresize.m
@@ -113,14 +113,14 @@ static void minMaxAutoLayoutSizes(NSWindow *w, NSSize *min, NSSize *max)
 	cw = uiprivMkConstraint(contentView, NSLayoutAttributeWidth,
 		NSLayoutRelationEqual,
 		nil, NSLayoutAttributeNotAnAttribute,
-		0, CGFLOAT_MAX,
+		0, 100000,
 		@"window maximum width finding constraint");
 	[cw setPriority:NSLayoutPriorityDragThatCanResizeWindow];
 	[contentView addConstraint:cw];
 	ch = uiprivMkConstraint(contentView, NSLayoutAttributeHeight,
 		NSLayoutRelationEqual,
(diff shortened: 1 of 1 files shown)


mischnic:
https://github.com/andlabs/libui/blob/cda991b7e252874ce69ccdb7d1a40de49cee5839/darwin/winmoveresize.m#L111-L119

Is there a case where a window with an uiArea can't grow infinitely large? There are no "non-stretchy" uiArea with a fixed size.

andlabs:
Not yet, but there will be.

mischnic:
Could it (a max width/height) somehow occur with another control in the window?

bildschirmfoto 2018-08-03 um 14 16 22


So is 100000 as maximum contraint width/height fine for now, because there is no SOMETHING_MAX value that works?

andlabs:
The fact there is no such constant is the reason why I was using `CGFLOAT_MAX`. I suppose I could make `100000` a constant, but I still wonder why this fails now on Sierra.

mischnic:
The largest possible value that works is `4294967384.0`, which is UINT_MAX + 89.

mischnic:
What keeps this from getting merged? Even if this might not be a perfect solution (although there seems to be no case in which the maximum size constraint has any effect) , it's still better than the application crashing.

Issue #318, opened 2018-03-25
Master issue on image drawing and image controls


andlabs:
Common request.

This is dependent on #310, as the table branch has a lot of the groundwork done already. Once that's done, I'll start discussing and planning the API (as I need to go back through my notes and the OS API documentation and also make sure I'm doing high-DPI multi-resource correctly).

This issue will be locked until existing issues are merged.

andlabs:
I'm stretching these master issues quite thin, it seems...

PR #386, opened 2018-06-01
melonDS: uiDrawBitmap

msink:


mischnic:
Related: #209, #318

andlabs:
Yes; this one might have to be rejected after uiImage gets merged in.

msink:
I'm not sure this can be replaced by uiImage - this one intended to *fast* drawing for games, with hardware acceleration if possible.

andlabs:
Yeah, I'll have to take that into consideration too. Probably a no-copy option will be needed, depending on what I can guarantee about pixel order.

OpenGL areas with the appropriate fragment shader might be better for this.

Issue #344, opened 2018-04-22
ScrollingArea setSize does nothing on Windows


mischnic:
It seems like `uiAreaSetSize` has no effect whatsoever on the area (at least on macOS and Windows 10):

```c
#include <stdio.h>
#include <string.h>
#include "ui.h"

double width = 0;
double height = 0;

static void handlerDraw(uiAreaHandler *a, uiArea *area, uiAreaDrawParams *p)
{
	uiDrawBrush brush;
	brush.Type = uiDrawBrushTypeSolid;
	brush.R = 255;
	brush.G = 0;
	brush.B = 0;
	brush.A = 1.0;

	uiDrawPath *path = uiDrawNewPath(uiDrawFillModeWinding);
	uiDrawPathAddRectangle(path, 0, 0, 1000, 1000);
	uiDrawPathEnd(path);	

	uiDrawFill(p->Context, path, &brush);
	uiDrawFreePath(path);
}

static void handlerMouseEvent(uiAreaHandler *a, uiArea *area, uiAreaMouseEvent *e) {
	width = 2*e->X;
	height = 2*e->Y;
	printf("%d %d\n", (int)width, (int)height);
	uiAreaSetSize(area, width, height);
	uiAreaQueueRedrawAll(area);
}
static void handlerMouseCrossed(uiAreaHandler *ah, uiArea *a, int left) {}
static void handlerDragBroken(uiAreaHandler *ah, uiArea *a) {}
static int handlerKeyEvent(uiAreaHandler *ah, uiArea *a, uiAreaKeyEvent *e) {return 0;}


int main(void)
{
	uiInitOptions o;
	const char *err;

	memset(&o, 0, sizeof (uiInitOptions));
	err = uiInit(&o);
	if (err != NULL) {
		fprintf(stderr, "error initializing ui: %s\n", err);
		uiFreeInitError(err);
		return 1;
	}

	uiWindow *mainwin = uiNewWindow("test", 600, 600, 1);
	uiAreaHandler handler;

	handler.Draw = handlerDraw;
	handler.MouseEvent = handlerMouseEvent;
	handler.MouseCrossed = handlerMouseCrossed;
	handler.DragBroken = handlerDragBroken;
	handler.KeyEvent = handlerKeyEvent;

	uiArea *area = uiNewScrollingArea(&handler, 400, 400);

	uiWindowSetChild(mainwin, uiControl(area));

	uiControlShow(uiControl(mainwin));

	uiMain();
	return 0;
}
```

[...]

PR #387, opened 2018-06-01
Fix area setSize macOS

mischnic:
Handles the macOS issues of #344.


andlabs:
Ah yes, this is correct, because our scrollview code no longer uses Auto Layout. In fact, the `invalidateIntrinsicContentSize` could go away too. I'll do that myself.

Issue #390, opened 2018-06-08
Memory Management


gnarz:
Hi there,
I made a binding of libui to a garbage collected language, and have hit a problem. Currently all container controls destroy their children when they are destroyed. However, in garbage collected languages, this violates the principle of least surprise. You do normally not expect an object to become invalid as long as you have a reference to it. What's more, you can't even tell that the object has become invalid, but the problem here is the object becoming invalid in the first place.

My current solution for this is to remove all children from container controls before I destroy them, but this does not work for uiGrid, which does not have a function to remov children, and menus in their current form are even more of a special case.

I would wish for all container widgets to have a method to delete all of their children (without destroying them, of course), or at least a method to delete individual widgets, paired with a way to actually find them. Like, the uiBox control has no way to determine the number of children, but the uiTab control has... plus a solution for menus, that addresses this problem.

mischnic:
@parro-it How is this handled in libui-node?

parro-it:
Hi @gnarz, we effectively  already solved the same problem for libui-node. I'm on mobile Atm, will give you more details later...

tom-corwin:
I +1 the question about managing the GC, since I know I'm not doing it right in my bindings either.

parro-it:
Well, like you probably already know, a `libui` control is represented in code by an instance of struct uiControl:

```C
typedef struct uiControl uiControl;

struct uiControl {
	uint32_t Signature;
	uint32_t OSSignature;
	uint32_t TypeSignature;
	void (*Destroy)(uiControl *);
	uintptr_t (*Handle)(uiControl *);
	uiControl *(*Parent)(uiControl *);
	void (*SetParent)(uiControl *, uiControl *);
	int (*Toplevel)(uiControl *);
	int (*Visible)(uiControl *);
	void (*Show)(uiControl *);
	void (*Hide)(uiControl *);
	int (*Enabled)(uiControl *);
	void (*Enable)(uiControl *);
	void (*Disable)(uiControl *);
};
```

One of it's members is `void (*Destroy)(uiControl *);`. libui set this member on control creation to a specific function for every control. Normal controls get a standard function, while containers get a more specialized one, that take cares of destroying all their children.

`libui` then calls this function on the window instance on window close, or on single widgets instance if they are removed from their containers early.:

https://github.com/andlabs/libui/blob/9cf6c3faf560d0ab69a46a540780ae405323dde8/common/control.c#L5-L8

...

parro-it:
[...]

PR #392, opened 2018-06-13
Add `void *user` into `struct uiControl`

msink:
This pointer dedicated to user program, libui should not touch it in any way.
If available, user program (or language binder) can avoid many ugly and dirty tricks.
8 bytes for each control is small cost for that flexibility.

### DIFF
--- a/ui.h
+++ b/ui.h
@@ -80,6 +80,7 @@ struct uiControl {
 	uint32_t Signature;
 	uint32_t OSSignature;
 	uint32_t TypeSignature;
+	void *user;
 	void (*Destroy)(uiControl *);
 	uintptr_t (*Handle)(uiControl *);
 	uiControl *(*Parent)(uiControl *);


andlabs:
Which dirty tricks (besides the overriding uiControlDestroy one in #390)?

mischnic:
> Which dirty tricks (besides the overriding uiControlDestroy one in #390)?

I guess using a map: https://github.com/andlabs/libui/issues/390#issuecomment-396011535

> > We also stored the libui control and the wrapper in a global map
>
> Maybe ask @andlabs to add void *data into uiControl struct?
>Retrieving from global map is O(N) operation, while with pointer in Control itself it will be O(1).

msink:
Well, for example for `uiAreaHandler` I currently have to append my `void *` after it:

```
struct ktAreaHandler {
    uiAreaHandler ui;
    void *user;
};
```

If it will in uiArea (via uiControl) - this will be unneded.

parro-it:
I also thought a user member would be really useful for these two use cases. 
I was sure I already opened an issue asking for such member some time ago, but I cannot find that issue anymore, I should have dreamt of it ☺️

andlabs:
For uiAreaHandler you can just use a pointer to `struct ktAreaHandler` as a pointer to `uiAreaHandler` the way you have it set up. But I do wonder if there are better ways to do this.

For destruction purposes I have considered an OnDestroy event.

parro-it:
> For destruction purposes I have considered an OnDestroy event.

The new purposed member does not replace the Destroy member override, that should still be used to hook in the destroy process to clean up wrapper resource.

What it avoids is the need to use a map data structure to retrieve a wrapper instance given an uiControl instance, since we can store the wrapper inside the new user member. 
The map data structure could introduce a delay when there are many controls in a window.

PR #399, opened 2018-07-02
Keep in memory getSizing information with hMessageFont on windows

franko:
Avoid calling getSizing with hMessageFont over and over, just keep memory of the sizing information the first time the function is called.

The idea is that hMessageFont and their metric does not depend on the specific window's handle.

### DIFF
--- a/windows/init.cpp
+++ b/windows/init.cpp
@@ -6,6 +6,7 @@ HINSTANCE hInstance;
 int nCmdShow;
 
 HFONT hMessageFont;
+uiWindowsSizing messageFontSizing;
 
 // LONGTERM needed?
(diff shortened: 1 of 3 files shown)


andlabs:
Those values will, however, depend on the DPI, and libui aims to be per-window DPI independent soon. The current implementation allows this.

franko:
Ok, I didn't have in mind this scenario, ie different per-monitor DPI, I don't even know how it works :-)

So do you want to do something anyway or just stick with the current approach ?

andlabs:
I'm not sure yet; I'll decide when I get around to doing DPI independence.

franko:
Ok, I understand. You can close the merge request if you want.

Issue #394, opened 2018-06-19
Background text of attributed string does not work properly on High Sierra when it wraps to multiple lines.


parro-it:
When a fragment of an attributed string that has background attribute wraps to multiple lines,
the background attribute extends to the end of the last line containing the fragment.

The problem happens only on macOS High Sierra. macOS Sierra is not affected.
The GIF below is taken from the drawtext example:

![giu-19-2018 21-57-44](https://user-images.githubusercontent.com/11197111/41621820-1af38154-740e-11e8-8128-aed1af51f163.gif)

We discover the problem on libui-node https://github.com/parro-it/libui-node/issues/101

mischnic:
Could this be a bug in High Sierra because the background is drawn by Core Text and not libui itself?

https://github.com/andlabs/libui/blob/cda991b7e252874ce69ccdb7d1a40de49cee5839/darwin/attrstr.m#L298-L309

Also, how does it behave on 10.11 and earlier?

https://github.com/andlabs/libui/blob/cda991b7e252874ce69ccdb7d1a40de49cee5839/darwin/future.m#L11-L12

PR #404, opened 2018-07-31
Custom drawing macOS BackgroundAttribute

mischnic:
Completes the text background drawing fallback if `kCTBackgroundColorAttributeName` is not available.

The output is exactly the same as `kCTBackgroundColorAttributeName`, apart from bug #394 so I would make this the default background drawing on High Sierra (and newer - does someone use Mojave?) and keep using `kCTBackgroundColorAttributeName` on older versions because it might be faster(?).

### DIFF
--- a/darwin/attrstr.h
+++ b/darwin/attrstr.h
@@ -89,3 +89,5 @@ extern CFAttributedStringRef uiprivAttributedStringToCFAttributedString(uiDrawTe
 - (id)initWithStart:(size_t)s end:(size_t)e r:(double)red g:(double)green b:(double)blue a:(double)alpha;
 - (void)draw:(CGContextRef)c layout:(uiDrawTextLayout *)layout at:(double)x y:(double)y utf8Mapping:(const size_t *)u16tou8;
 @end
+
+CGColorRef mkcolor(double r, double g, double b, double a);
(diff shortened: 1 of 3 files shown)


andlabs:
I haven't actually integrated custom background blocks yet because to get the behavior I was looking for would require knowledge of per-line metrics, which is a massive mess of partially undocumented features and algorithms. I have to see how the attributes behave on all the versions of all the platforms I'm interested in.

mischnic:
> I haven't actually integrated custom background blocks yet because to get the behavior I was looking for would require knowledge of per-line metrics

Are you referring to a background like this (the red part)?
bildschirmfoto 2018-07-31 um 15 11 45

That doesn't seem too complicated on the macOS side.

andlabs:
Yes. The question I was asking is which is correct: the dark red or the light red. If the behavior I want is the light red (and I do want that behavior for text selections, at the least), I'd need to reinstate the old metrics code.

andlabs:
My point is I'm not sure which I want; I have to double-check with all platforms to see which options I do have. If I have both options after all, I'd go with the former.

mischnic:
Current situation on GTK:

bildschirmfoto 2018-08-01 um 12 58 53

PR #405, opened 2018-08-06
OpenGL control

mischnic:
I tried to update @pcwalton's work onto master.

Working fine on macOS:
bildschirmfoto 2018-08-06 um 23 11 29

I can't get it work on linux:
- Nothing is drawn (not even the black background) - I'm no OpenGL expert all
- Mouse events don't work (@andlabs this is basically all of uiArea's code but `motion_notify_event` and others don't get called at all ??)

Windows is still missing.

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -7,7 +7,7 @@ list(APPEND _LIBUI_SOURCES
 	common/areaevents.c
 	common/control.c
 	common/debug.c
-#	common/drawtext.c
+	common/opengl.c
 	common/matrix.c
 	common/opentype.c
 	common/shouldquit.c
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -1,14 +1,14 @@
 # 3 june 2016
 
 list(APPEND _LIBUI_SOURCES
-	common/areaevents.c
 	common/attribute.c
 	common/attrlist.c
 	common/attrstr.c
+	common/areaevents.c
 	common/control.c
 	common/debug.c
+#	common/drawtext.c
 	common/matrix.c
-	common/opengl.c
 	common/opentype.c
 	common/shouldquit.c
 	common/userbugs.c
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -1,14 +1,14 @@
 # 3 june 2016
 
 list(APPEND _LIBUI_SOURCES
+	common/areaevents.c
 	common/attribute.c
 	common/attrlist.c
 	common/attrstr.c
-	common/areaevents.c
 	common/control.c
 	common/debug.c
-#	common/drawtext.c
 	common/matrix.c
+	common/opengl.c
 	common/opentype.c
 	common/shouldquit.c
 	common/userbugs.c
--- a/common/controlsigs.h
+++ b/common/controlsigs.h
@@ -17,6 +17,7 @@
 #define uiGroupSignature 0x47727062
 #define uiLabelSignature 0x4C61626C
 #define uiMultilineEntrySignature 0x4D6C6E45
+#define uiOpenGLAreaSignature 0x4F474C41
 #define uiProgressBarSignature 0x50426172
 #define uiRadioButtonsSignature 0x5264696F
 #define uiSeparatorSignature 0x53657061
--- a/common/opengl.c
+++ b/common/opengl.c
@@ -1,0 +1,90 @@
+// 28 may 2016
+
+#include <stdlib.h>
+#include "../ui.h"
+#include "uipriv.h"
+
+uiOpenGLAttributes *uiNewOpenGLAttributes() {
+    uiOpenGLAttributes *a = malloc(sizeof(uiOpenGLAttributes));
+    if (a == NULL)
+        return NULL;
+
+    a->RedBits = a->GreenBits = a->BlueBits = a->AlphaBits = 8;
+    a->DepthBits = 24;
+    a->StencilBits = 8;
+    a->Stereo = 0;
+    a->SRGBCapable = 0;
+    a->DoubleBuffer = 1;
+    a->UseOpenGLES = 0;
+    a->MajorVersion = 1;
+    a->MinorVersion = 0;
+    a->ForwardCompat = 0;
+    a->DebugContext = 0;
+    a->CompatProfile = uiOpenGLDontCare;
+    a->Robustness = 0;
+
+    return a;
+}
+
+void uiFreeOpenGLAttributes(uiOpenGLAttributes *attribs) {
+    free(attribs);
+}
+
+void uiOpenGLAttributesSetAttribute(uiOpenGLAttributes *attribs, uiOpenGLAttribute attribute, int value) {
+    switch (attribute) {
+    case uiOpenGLAttributeRedBits:
+        attribs->RedBits = value;
+        break;
+    case uiOpenGLAttributeGreenBits:
+        attribs->GreenBits = value;
+        break;
+    case uiOpenGLAttributeBlueBits:
+        attribs->BlueBits = value;
+        break;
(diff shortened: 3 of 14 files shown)


andlabs:
Try putting the area in a box and marking it stretchy, then adding a button. There's glitches involving the sizing of areas that are by themselves in windows on GTK+.

andlabs:
Did that work?

mischnic:
> Did that work?

Yes!

mischnic:
The OpenGL error is 0x502 (GL_INVALID_OPERATION?) in these two lines, but not on macOS:
```c
glUniformMatrix4fv(openGLState.ProjectionUniform, 1, GL_FALSE, &projection.m11);
glUniformMatrix4fv(openGLState.ModelViewUniform, 1, GL_FALSE, &modelview.m11);
```

mischnic:
@andlabs Is the macOS area refactoring acceptable? `areaCommonView` now handles all events and is the superclass of `openGLAreaView` and `areaView` .

mischnic:
Do we want a scrolling opengl view?

mischnic:
@andlabs For linux and macOS, this code works because the constructor runs synchronously and the context is actually created (see the opengl example):
```c
uiOpenGLArea *glarea = uiNewOpenGLArea(&AREA_HANDLER, attribs);
uiOpenGLAreaMakeCurrent(glarea);

GLenum glewStatus = glewInit();
if (glewStatus != GLEW_OK) {
	fprintf(stderr, "GLEW init error: %s\n", glewGetErrorString(glewStatus));
	exit(EXIT_FAILURE);
}

// ...

uiMain();
```

On Windows however, the context creating happens in `areaWndProc` (see openglarea.c) so `uiOpenGLAreaMakeCurrent` tries to make `NULL` current (this needs to be done before `glewInit` can run). How can this be done?

mischnic:
On Windows, the triangle is only visible for a short duration after rendering, then the area goes completely black (setting a background color using `glClearColor` works correctly).

mischnic:
@andlabs The OpenGL area is now working on every platform, even with OpenGL 3.3. Only some cleanup and  better openglattributes handling needs to be done. Here are a few architectural questions I have:

- Does including x11 libraries (for GLX) always use XWayland, is there a difference to before? The native wayland interface is EGL.
- How to do error handling (e.g. if requested attributes like OpenGL version area not available):
	- return an error code and add an enum for these OR return NULL and pass an error string in some other way
	- throwing a uiprivUserbug is non recoverable so a program can't even display "Your graphics card is not supported".
- Do you want scrolling views?
- Why can't the 32-bit Linux Travis find `-lGL`.

cody271:
@mischnic Instead of using GLX / X11 APIs directly in a custom widget, we should probably be using the included [`GtkGLArea`](https://developer.gnome.org/gtk3/stable/GtkGLArea.html) widget. (GTK 3.16 added this, over three years ago now..)

andlabs:
I would need to bump the minimum version for that. Not sure if I should or not. Again, will look at all this soon and see if there's a better approach if bumping the version isn't suitable.

mischnic:
Instead of GLX, EGL could be used on Wayland. But I dont know how to detect this at runtime and if these libraries would need to be loaded dynamically.

ThomasCassimon:
Hi, I was wondering what the current state of this feature is?

The [3.0 changelog](https://github.com/andlabs/libui/blob/master/NEWS.md#old-news) seems to indicate this was planned for the 4.0 release, but I don't believe this feature has made it in yet, has it?

CI seems to be [failing for 32-bit Linux](https://travis-ci.org/andlabs/libui/jobs/423707135), because it can't find "-lGL", (the GL library), but when looking at the [CMakeLists.txt](https://github.com/andlabs/libui/blob/31ec22b846e020250de0909cc4ae1e3e13c5dede/examples/CMakeLists.txt#L65) I was wondering why you are manually specifying the library names instead of using [FindOpenGL](https://cmake.org/cmake/help/latest/module/FindOpenGL.html)?

mischnic:
> Hi, I was wondering what the current state of this feature is?

Waiting for @andlabs' feedback/review.

> I was wondering why you are manually specifying the library names instead of using FindOpenGL?

Fixed. 😄 Let's see what happens.

mischnic:
@andlabs Why isn't Travis running? For #424 neither.

mischnic:
@andlabs ?

andlabs:
I don't know; that's weird. Travis still lets me see the status of libui, but it's behind...

mischnic:
Now the linux 32bit Travis job just hangs while downloading packages: https://travis-ci.org/andlabs/libui/jobs/438028266

andlabs:
Okay I guess I needed to just log in again and it seems to be back.

I should probably get to updating both the issues and pull requests for this now (I was taking a break after releasing Alpha 4 and package `ui`, but IDK how long that will continue for)

paulrouget:
The Mac example doesn't start rendering until the window is resized, and then, it flickers a lot.

Any idea what's going on?

mischnic:
> The Mac example doesn't start rendering until the window is resized

Could you please check if `onDrawGL` is called before resizing and if it is, whether the correct width & height (= non-zero) values are passed?

```diff
diff --git a/examples/opengl/main.c b/examples/opengl/main.c
index 1fff59c9..ab19e5a7 100644
--- a/examples/opengl/main.c
+++ b/examples/opengl/main.c
@@ -277,6 +277,7 @@ static void onInitGL(uiOpenGLAreaHandler *h, uiOpenGLArea *a)
 static void onDrawGL(uiOpenGLAreaHandler *h, uiOpenGLArea *a, double width, double height)
 {
        GLCall(glViewport(0, 0, width, height));
+       printf("%f %f\n", width, height);

        GLCall(glClear(GL_COLOR_BUFFER_BIT));
```

Should print: `600.000000 415.000000`

> and then, it flickers a lot.

Maybe disabling VSync (`uiOpenGLAreaSetVSync(a, 0);` in setup) helps?
Is there a difference with this change? (Still flickering when moving the mouse over the area?)
```diff
diff --git a/examples/opengl/main.c b/examples/opengl/main.c
index 1fff59c9..5e1c9914 100644
--- a/examples/opengl/main.c
+++ b/examples/opengl/main.c
@@ -362,8 +363,6 @@ int main(void)
        uiOpenGLArea *glarea = uiNewOpenGLArea(&AREA_HANDLER, attribs);
        uiBoxAppend(b, uiControl(glarea), 1);

-       uiTimer(1000/60, render, glarea);
-
        uiControlShow(uiControl(mainwin));
        uiMain();
```


<br>

Works fine for me (macOS 10.13.6, Intel Graphics).

paulrouget:
onDrawGL is called before resizing.

It's printing the right size I believe (600x416).

Disabling vsync doesn't change anything.

Still flickers without the timer.

I think every odd frame is transparent:

Screencast: https://streamable.com/12xcn

paulrouget:
> I guess you're running Mojave?

Yes. 10.14.1.

> This should definitely fix the blank-before-resize-issue:
> [self->libui_a->ctx setView:self];
> [self->libui_a->ctx update];

This change changes nothing.

> When setting `wantsLayer` to true, the flickering happens for me as well, please try this change

Setting `setWantsLayer` to `NO` changes nothing.

Setting `setWantsLayer` to `YES` fixes the blank-before-resize-issue.

No matter the configuration/changes, it's flickering.

mischnic:
`setWantsLayer:NO` on Mojave: Blank before resize and flickering on every other frame
`setWantsLayer:NO` on High Sierra: Works fine
`setWantsLayer:YES` on Mojave: Flickering
`setWantsLayer:YES` on High Sierra: Flickering

@pcwalton You have investigated this some time ago, any suggestion?

Shadowblitz16:
Will this work on mac os?
I know opengl is deprecated on mac os.

PR #413, opened 2018-08-17
Table - support selection access

bcampbell:
This stuff adds support for:
- an OnSelectionChanged callback
- uiTableCurrentSelection() to read the currently-selected items
- a MultiSelect flag (to let the user select more than one item at a time)

There's also a table example, which I'll keep in a separate branch for now: see https://github.com/bcampbell/libui/tree/table-example

Just unix (Gtk) version implemented so far. I've got my old code for windows and OSX more or less ready too, but I'll hold off until the API is nailed down.
(My older version returned an iterator object instead of `uiTableSelection`. It was fiddlier to use, but there might still be a good argument for more abstract access).

### DIFF
--- a/ui.h
+++ b/ui.h
@@ -1463,6 +1463,9 @@ _UI_EXTERN void uiTableAppendButtonColumn(uiTable *t,
 _UI_EXTERN void uiTableOnSelectionChanged(uiTable *t, void (*f)(uiTable *t, void *data), void *data);
 
 // uiTableSelection holds an array of row indexes for a table.
+// it's safe to fiddle with the Items data in place (eg a caller
+// might want to sort them - that's ok). But probably best not to
+// change NumItems or try to reallocate Items.
 typedef struct uiTableSelection uiTableSelection;
 struct uiTableSelection
 {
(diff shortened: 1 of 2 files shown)


bcampbell:
(just to be clear - I don't expect this to be merged in without windows and OSX support. But got to start somewhere :- )

msink:
Why cant that `uiTableSelection` be just one more column in TableModel?

andlabs:
Because different tables can have different selections but share the same model.

msink:
Sure, but same model does not mean same data.
Still dont see difference...

bcampbell:
The selection is a property of the view (ie the table widget) rather than the underlying data (the model).
So "same model" _does_ mean "same data", I think.
You might reuse the same `uiTableModelHandler` functions for more than one model (although looking at it now, you'd probably want a userdata field on `uiTableModel` to facilitate this...)

gnarz:
@bcampbell I did propose userdata for table models in #414

szanni:
@bcampbell did you ever get around to implementing darwin & windows fully?

bcampbell:
> @bcampbell did you ever get around to implementing darwin & windows fully?

Sorry, no. In any case, I no longer have access to a mac :-(

szanni:
> Sorry, no. In any case, I no longer have access to a mac :-(

Sad to hear, but thank you for the quick and prompt response! I guess I'll take a shot at this myself then :)

PR #424, opened 2018-09-26
Make slider tooltip default on all platforms. Add slider tooltips with enable/disable functions.

szanni:
This is an updated version of #380 addressing the concerns raised in that pull request.
As can be seen in the README the slider is not consistent across platforms.
This set of patches disables the value drawn over the slider on unix platforms and replaces it with a tooltip. macOS has a tooltip now by default too.
The displaying of tooltips can be enabled/disabled via the uiSliderSetHasTooltip function.
uiSliderHasTooltip can be used to check the current state of the slider tooltip.

Tested on Linux, macOS and wine/mingw-w64.


mischnic:
*Ideally*, the tooltip could be set by the user (via a callback for a different range or specifying units), but being able to disable them at least makes it possible to avoid confusion if two different values are displayed (e.g. in the tooltip and a separate label).

szanni:
> _Ideally_, the tooltip could be set by the user (via a callback for a different range or specifying units), but being able to disable them at least makes it possible to avoid confusion if two different values are displayed (e.g. in the tooltip and a separate label).

I thought about that. Unix and macOS should be fairly easy in that regard, but windows seems very convoluted. Or well, the MSDN docs are just crap. I've been browsing the wine source code to get a grip on tooltips but I haven't quite figured out how to set custom messages. It looks like _TTM_UPDATETIPTEXT_ would be the message to send on every value change!? Since that takes a WCHAR every call would entail a conversion as well? I definitely want to look into that when I have more time. But I would think of this being a different API call.

PR #434, opened 2018-10-29
examples/cpp-multithreaded: fixed locks and timeThread terminate conditions

marioc64:
C++ unique_lock created through constructor with an argument creates already locked lock. The lock is unlocked with destructor.

condition_variable::wait_for returns true when await at least specified timeout false otherwise. There is no guarantee that it returns false when notified, therefore thread exit is done through extra flag 'running'.

### DIFF
--- a/examples/cpp-multithread/main.cpp
+++ b/examples/cpp-multithread/main.cpp
@@ -12,8 +12,8 @@ using namespace std;
 uiMultilineEntry *e;
 condition_variable cv;
 mutex m;
-unique_lock<mutex> ourlock(m);
 thread *timeThread;
+volatile bool running = true;
 
 void sayTime(void *data)
 {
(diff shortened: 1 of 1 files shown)


examples/cpp-multithread/main.cpp
```diff
@@ -12,8 +12,8 @@ using namespace std;
 uiMultilineEntry *e;
 condition_variable cv;
 mutex m;
-unique_lock<mutex> ourlock(m);
 thread *timeThread;
+volatile bool running = true;
```
paulsapps: this should be std::atomic<bool>

marioc64:
There is something wrong with one test build:

> Pipeline configuration error:
/azure-pipelines.yml: Could not find /azure-pipelines/artifacts.yml in repository self hosted on https://api.github.com using commit 05c688b52ba089b453ea05b3104a9116e6f10544. GitHub reported the error, "Not Found"

PR #440, opened 2018-11-28
Improved multi-line entry on macOS

frarees:
- Proper bg and fg colors on Dark Mode (Mojave)
- No use of NSRegularControlSize (deprecated 10.12)
- Keeps styling between typed characters and scripted characters
(SetEntry/Append)

Mojave (Dark) without patch:
mojave_wofix

Mojave (Dark) with patch:
mojave_wfix

High Sierra (compiled in Mojave) with patch:
highsierra_wfix

### DIFF
--- a/darwin/multilineentry.m
+++ b/darwin/multilineentry.m
@@ -6,6 +6,8 @@
 // see also http://stackoverflow.com/questions/24210153/nstextview-not-properly-resizing-with-auto-layout and http://stackoverflow.com/questions/11237622/using-autolayout-with-expanding-nstextviews
 @interface intrinsicSizeTextView : NSTextView {
 	uiMultilineEntry *libui_e;
+	NSDictionary *strAttrs;
+	NSFont *font;
 }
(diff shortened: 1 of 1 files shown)


andlabs:
NSRegularControlSize must still be used to compile on 10.8 - 10.10. Will look in more detail later.

frarees:
> NSRegularControlSize must still be used to compile on 10.8 - 10.10.

I'm using `NSFont controlContentFont` ([ref](https://developer.apple.com/documentation/appkit/nsfont/1527070-controlcontentfont)):

> If fontSize is 0 or negative, returns the control content font at the default size.

It's been there since 10.0. I'm not sure if this overload of the specific usage of `0` here is something introduced lately and it's hard to check without a setup, but I've also found [this](https://www.gnu.org/software/gnustep/resources/documentation/Developer/Gui/Reference/NSFont.html#method$NSFont+boldSystemFontOfSize$)

PR #448, opened 2018-12-31
Add cmake options: BUILD_TESTS & BUILD_EXAMPLES

solvingj:
These options are ubiquitous among many CMake projects. For many scenarios, building and running unit tests and examples on each build is unnecessary and/or undesirable.  This PR leaves the default behavior (both options set to `ON`), but lets users opt-out.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,8 @@ endif()
 
 project(libui)
 option(BUILD_SHARED_LIBS "Whether to build libui as a shared library or a static library" ON)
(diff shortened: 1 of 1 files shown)


andlabs:
I have decided to drop cmake; sorry.

solvingj:
since removal of cmake would take quite a lot of time, perhaps you would accept this trivial improvement for the meantime, and until it's totally phased out.

andlabs:
Actually I've just finished the replacement (though it won't be merged into master until after I set up a new CI environment), and the new setup has the unfortunate (for now) side effect that all options are built always. Perhaps this could be set as Meson options, but. Thanks anyway!

PR #452, opened 2019-01-17
Remove GTK timer source ID from main context on uninit

gdiv:
Restarting the main loop without removing the timer source results in
null-pointer dereference error in GTK.

https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html#mainloop-memory-management

### DIFF
--- a/unix/main.c
+++ b/unix/main.c
@@ -22,11 +22,17 @@ const char *uiInit(uiInitOptions *o)
 	return NULL;
 }
 
-struct timer;		// TODO get rid of forward declaration
+struct timer {
+	int (*f)(void *);
+	void *data;
+	guint gtk_source_id;
+};
 
 static void uninitTimer(gpointer key, gpointer value, gpointer data)
 {
-	uiprivFree((struct timer *) key);
+	struct timer *t = (struct timer *) key;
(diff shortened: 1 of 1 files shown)


ak-coram:
> What are you trying to do here? uiUninit() isn't supposed to be followed by anything else...

What's the purpose of `uiUninit()`then?

cody271:
This looks correct to me FWIW. I thought `uiUninit()` should consistently clean up all global state..

gdiv:
> What are you trying to do here? `uiUninit()` isn't supposed to be followed by anything else...

Two things.

1. proper cleanup (that's consistent with GTK documentation)
2. being able to start the whole UI again without killing the process which is helpful in dynamic languages like Python during dev time

cody271:
> being able to start the whole UI again without killing the process which is helpful in dynamic languages like Python during dev time

Very useful for implementing hot reload in language bindings! :+1:

andlabs:
Okay, hot-reloading would require a bit more work than just this, but this PR would be a start to that. I'm leaning toward letting this idea wait until after I do the Alpha 5 cleanups, but eh.

PR #454, opened 2019-02-01
Point to the correct canonical location for Rust

NoraCodes:
I changed my name (from LeoTindall) and moved my fork to an organization, so it should be updated in the README.


andlabs:
Good to know, thanks!

PR #480, opened 2019-12-29
Update README.md

rjopek:
Removed outdated link, original project restored.


andlabs:
Very well, thanks again! And yes I will continue to be working on this. Happy new year! (Consider this the reply to the email you sent as well.)

PR #481, opened 2020-01-12
Set appropriate colours for Multiline Entry control on dark mode (Darwin)

TropicalBastos:
The PR checks in the following:

- White font colour for multi line entry if user has dark mode enabled
- Dark background colour for multi line entry if user has dark mode enabled
- Utility method for checking if dark mode is enabled
- .gitignore entry for ignoring the build folder


TropicalBastos:
Only just noticed that there has been a previous PR for this particular dark mode issue. However looks like the previous one did not pass all tests.

LoganDark:
`isDarkMode` should be either be `uiprivIsDarkMode` and inside of `uipriv_darwin.h`, or `uiDarwinIsDarkMode` and inside of `ui_darwin.h` (honestly, the latter would be more useful because then people can use it in their libui apps).

PR #484, opened 2020-02-05
Fix #416, uiTableModelRowInserted on windows adds 2 rows instead of 1.

### LINKED ISSUE #416: uiTableModelInsertRow() causes invalid rows to be accessed (on windows)
If uiTableModelInsertRow() is called on an active uiTable, the table will then try to fetch one row more than what the model provides.
If the table already has a lot of content, the invalid fetch won't occur until you scroll right down to bring the last row into view.

Seems to be windows-specific.

example here:
https://github.com/bcampbell/libui/blob/destruc-table/examples/destruc-table/main.c

pftbest:
When LVM_INSERTITEM is called after we updated LVM_SETITEMCOUNT, the actual count will be increased by one causing one extra item to appear in the table and loading data from the model.
This extra item is invisible at first because it doesn't get painted but you can click on it to confirm that it is actually there.

To fix the problem we need to call INSERTITEM first and then update the ITEMCOUNT.

Fixes #416
Fixes andlabs/ui#369

cc @bcampbell
r? @andlabs

### DIFF
--- a/windows/table.cpp
+++ b/windows/table.cpp
@@ -36,16 +36,16 @@ void uiTableModelRowInserted(uiTableModel *m, int newIndex)
 	item.iItem = newIndex;
 	item.iSubItem = 0;
 	for (auto t : *(m->tables)) {
+		// update selection state
+		if (SendMessageW(t->hwnd, LVM_INSERTITEM, 0, (LPARAM) (&item)) == (LRESULT) (-1))
+			logLastError(L"error calling LVM_INSERTITEM in uiTableModelRowInserted() to update selection state");
+
 		// actually insert the rows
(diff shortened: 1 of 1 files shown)


szanni:
Patch looks good to me.

The MS docs don't mention anything about setting an item count on insertion.
Looking (again) at the WINE sources however tells me the list view increases its internal counter on `LVM_INSERTITEM` with `LVS_OWNERDATA` (which libUI uses).
Debug printing the item count before and after insertion confirms this.

My only question would be: why not remove the call to `LVM_SETITEMCOUNT` entirely? It is clearly redundant.

andlabs:
Yeah, I tried looking up why I did this, and couldn't find anything. "Update selection state" is a very vague comment in hindsight... I guess you can try removing the call and seeing what happens.

Issue #349, opened 2018-04-25
On enter event for uiEntry


e0ff:
I think these would be good events for uiEntry. On enter would be triggered after pressing the enter key.

bcampbell:
Doesn't `uiEntryOnChanged()` handle the update?

A while back I did a patch to add an event for when editing has 'finished' (not necessarily by pressing the enter key). It needs a little smoothing, but it works and might be useful for you:

https://github.com/andlabs/libui/pull/290

endoffile78:
Sorry i ditn't see that existed I'll edit the issue.

Issue #329, opened 2018-03-31
Master issue for missing features in uiArea


andlabs:
Merging a bunch of issues here.

TODO

TODO blah blah blah locked until merge

mischnic:
> but this must be clearly marked as only for the purposes of user interaction, NOT FOR TEXT INPUT (and I have to remember a good reason why; TODO)

Would shift + a key be properly parsed as well (e.g. capital letters, punctuation mark)?

andlabs:
It should, yes.

msink:
On Linux if Area set as direct child of Window - it is 1 pixel width, on Windows - fills entire Window.

mischnic:
- mousewheel/scroll event in non-scrolling areas
- keyevent with lower delay interval for games (noticed it on macOS)

AndyObtiva:
I noticed that when hooking a key event to area, and I press shift+1 to produce '!', the key value I receive from area key event is that of number 1 (49), not the '!' key value (33) produced by shift+1. This means that in order for me to recognize what character the user intended, I would have to access the current keyboard layout and perform a conversion, which is not something that is typically available in Ruby as far as I know. I think this must be done automatically by libui, giving me the key value for '!' not 1 when I press shift+1.

The SWT library handles this by offering an additional attribute "character" in its key event, which in the case of the example above is the calculated character from holding shift and pressing a key (they still provide the standard key value as well, meaning for shift+1, I get key value as 49 and character as 33 in SWT)

This would be great to add to facilitate fully manual manipulation of user key input should I want to build a string from user input.

Thanks.

PR #488, opened 2020-03-16
entry. handle keyboard events to then reuse in go binding

RustamGamidov:
this idea is quite popular basing on the feature-requests.
related to #290 #349 #329


RustamGamidov:
@andlabs could you have a look please?

PR #493, opened 2020-03-25
button: add uiButtonSetMinSize to avoid very small buttons

RustamGamidov:
it looks not very nice for example in the OK - Cancel dialogues

### DIFF
--- a/common/general.h
+++ b/common/general.h
@@ -1,0 +1,9 @@
+#ifndef H_COMMON_GENERAL
+#define H_COMMON_GENERAL
+
+int max(int first, int second)
+{
+	return (first < second) ? second : first;
+}
+
+#endif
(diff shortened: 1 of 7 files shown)


RustamGamidov:
macos ci is failing at the checkout statge.
ci UI is different from the one ~one week ago so this could be the reason

andlabs:
Hi. You do not need to keep closing and re-opening new pull requests. I will process this when I can, but I did see it the first time. Thanks in the meantime!

RustamGamidov:
sorry for the noise. it's because of nominmax in the MSVS

andlabs:
Oh, I see — no, you do not need to file a new PR every time you make changes. Just upload to your forked branch and github will pick up the changes.

RustamGamidov:
github is updating opened PR with the changes pushed to the original branch.
Sometimes there are regexp masks for the PR titles to skip CI like [WIP] or so, not sure if it's the case.

And i'm using mingw locally so everything was fine for me (as it usually is :) )
Really sorry about all that extra builds

andlabs:
Huh, that's new to me, and it would definitely inhibit contribution. I'll need to look into that. Sorry for snapping earlier!

PR #500, opened 2020-06-07
Fix memory leak in uiTable setEditable on unix.

L4r0x:
The GtkTreePath was not freed correctly.

I found this memory leak with Valgrind on ubuntu 19 after noticing that memory consumption slowly increases in the *test* application.


andlabs:
Good catch; thanks! I've fixed the CI failing; if you'd like to see it pass then you'll need to sync with my master

PR #507, opened 2020-08-25
Fix table editing redraw code on all platforms.

szanni:
Pretty much all information in the commit message. No
uiTableModelRowChanged was emitted on any of the changes when
editing cells in uiTable:

This fixes bugs on all platforms not calling uiTableModelRowChanged()
when setting a new value in edit mode. This is now automatically
done in uiprivTableModelSetValue() so that ALL uiTable views are
informed about the update.

Darwin and windows did some custom redrawing which hid this bug.
Unix does frequent unrelated redraw which hide the bug. It can often
be experienced when double clicking a checkbox.

### DIFF
--- a/common/tablemodel.c
+++ b/common/tablemodel.c
@@ -41,6 +41,14 @@ void uiprivTableModelSetCellValue(uiTableModel *m, int row, int column, const ui
 
 	mh = uiprivTableModelHandler(m);
 	(*(mh->SetCellValue))(mh, m, row, column, value);
+
+	// Abort redraw for button columns which signal being clicked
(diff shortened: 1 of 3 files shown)


szanni:
OK, I updated the PR removing the early abort to not trigger an `uiTableModelRowChanged` call when setting `NULL`.
Not only do button columns use a NULL value, so do color columns... as I found out doing some further code digging.

IMO we should call a `uiTableModelRowChanged` in any case **because we are setting a value** so the row technically changed!
API abuses to trigger other functionality should be fixed at its root. For example providing a callback handler for a button column or something.

PR #517, opened 2020-09-29
docs: Fix simple typo, plaform -> platform

timgates42:
There is a small typo in windows/_rc2bin/winapi.hpp, windows/winapi.hpp.

Should read `platform` rather than `plaform`.


andlabs:
Thanks!

